PR #55, opened 2023-11-21
no-issue - Refacto MakinaCorpus\DbToolsBundle\Anonimizer namespace

SimonMellerin:


pounard:
I'm not sure this was 100% necessary, but I like the new organization with config in its own namespace.

PR #56, opened 2023-11-22
A better dev script

SimonMellerin:


dev.sh
```diff
@@ -0,0 +1,135 @@
...
+    # Show autodoc help
+    awk '{ if ($0 ~ /^#[^!]/) { \
+                gsub(/^#/, "", $0); print $0 } }' "$0"
+    echo " "
+}
+
+args=${@:-usage}
+action=${1-}
+
+if [[ -n $@ ]];then shift;fi
```
SimonMellerin, line 130: @pounard  here we "shift " given arguments, here the "action", so after this line, `$@` contains agruments `$2...$n`

Issue #12, opened 2023-11-21
DbToolsStorage filename strategy


SimonMellerin:

jrmgx:
I'm interested in this feature, is it implemented? I'm a bit lost because it's marked as closed but no other information. Let me know :)

pounard:
This was closed because we were targeting an MVP and didn't want to pollute our task list with potentially useless features. We wanted some user feedback in order to priorize tasks and features.

Originally, the idea of having a file strategy was to allow users to use their own directory structure for dumps, along some default implementations, such as `<YEAR>/<MONTH>/` for example.

Further we'd like to have this hidden behind an `FilenameStrategyInterface` or such for allowing users to implement their own.

Is that what you'd need ?

jrmgx:
I understand now, it's exactly what I need.
The use case is: I have a project with a global backup mechanism in place, but it needs the file to follow a specific naming convention.

pounard:
Understood ! That's indeed a use case we would want to support as well.

OK, I leave this issue opened, we should be able to work on this within the next week, please do not hesitate asking for a status update.

pounard:
@jrmgx feature is now implemented and will be part of the next release.

In order to use it, you need to ether extend `MakinaCorpus\DbToolsBundle\Storage\AbstractFilenameStrategy` if don't need to change the root storage directory, or implement `MakinaCorpus\DbToolsBundle\Storage\FilenameStrategyInterface` if you need it to be different (if you have a single database, you don't need to fully implement the interface, but you may simply change the default storage path in configuration instead).

One important thing, the final directory will always be the concatenation of the `getRootDir()` and the `generateFilename()` methods. Default `getRootDir()` implementations suffices the root directory with the connection name as a folder, if you want to change this behavior you need to implement fully the interface.

Once done, you need to register your custom implementation using this:

```yaml
# file: config/packages/db_tools.yaml
db_tools:
    storage:
        filename_strategy:
            my_dbal_connection_name: App\DbTools\Storage\MyCustomFilenameStrategy
```

The strategy can be a stateless class with no constructor arguments, case in which you don't need to register it into the container and let the extension will do it for you, or it be configured using the container, case in which you can replace the class name by the service identifier in configuration (since it's the best practice to leave the class name be the service identifier, it should not change the configuration in the end, but if you have multiple connections and need to register more than once the same class, you can name the services, it'll work).

PR #81, opened 2024-02-12
issue #12 - filename strategy

pounard:
Adds an interface:

```php
interface FilenameStrategyInterface
{
    /**
     * Generate dump file name.
     */
    public function generateFilename(
        string $storageRootDir,
        string $connectionName = 'default',
        string $extension = 'sql',
        bool $anonymized = false
    ): string;
}
```

Configuration goes like this:

```yaml
db_tools:
    # ...

    connections:
        default:
            filename_strategy: MakinaCorpus\DbToolsBundle\Filename\StupideTestingFilenameStrategy
        foo_connection:
            filename_strategy: "default" # can be "datetime"
        mysql:
            filename_strategy: ~

```

And documentation is missing, do not merge yet !

@SimonMellerin any opinions ?


config/packages/db_tools.yaml
```diff
@@ -44,3 +44,22 @@ db_tools:
         #yaml:
             #- connection_one: '%kernel.project_dir%/config/anonymizations/connection_one.yaml'
             #- connection_two: '%kernel.project_dir%/config/anonymizations/connection_two.yaml'
+
+    # Connection-specific configuration.
+    # Keys are DBAL connection names registered in container configurfation.
+    #connections:
+        #default:
```
SimonMellerin, line 51: I find a bit odd this key `connections`, I'd rather see a dedicated section like `storage` as first level key

src/Filename/DefaultFilenameStrategy.php
```diff
@@ -0,0 +1,34 @@
+<?php
+
+declare(strict_types=1);
+
+namespace MakinaCorpus\DbToolsBundle\Filename;
```
SimonMellerin, line 5: May be we should add an  `MakinaCorpus\DbToolsBundle\Storage` namespace and put the `DbToolsStorage` class in it an rename it `Storage` + these `FilenameStrategy` classes?

SimonMellerin (changes requested):

docs/content/configuration.md
```diff
@@ -13,7 +13,7 @@ Some options are available to customize how the `db-tools:backup` command works.
 
 ### Storage directory
```
SimonMellerin, line 14: Maybe here you can rename with

```suggestion
### Storage 
```

And explain a bit that one can implement its own Filename strategy

src/DependencyInjection/DbToolsExtension.php
```diff
@@ -23,7 +26,10 @@ public function load(array $configs, ContainerBuilder $container)
         $loader = new YamlFileLoader($container, new FileLocator(\dirname(__DIR__).'/../config'));
         $loader->load('services.yaml');
 
-        $container->setParameter('db_tools.storage_directory', $config['storage_directory']);
+        if (isset($config['storage_directory'])) {
+            \trigger_deprecation('makinacorpus/db-tools-bundle', '1.0.1', '"db_tools.storage_directory" configuration option is deprecated and renamed "db_tools.storage.root_dir"');
+        }
```
SimonMellerin, line 31: I don't know if this is necessary as you set this configuration as deprecated in `DbToolsConfiguration.php`.

But I guess we can keep it

src/DependencyInjection/DbToolsExtension.php
```diff
@@ -50,6 +56,30 @@ public function load(array $configs, ContainerBuilder $container)
         $anonymizerPaths[] = \realpath(\dirname(__DIR__)) . '/Anonymization/Anonymizer';
 
         $container->setParameter('db_tools.anonymization.anonymizer.paths', $anonymizerPaths);
+
+        // Register filename strategies.
+        $strategies = [];
+        foreach (($config['storage']['filename_strategy'] ?? []) as $name => $strategyId) {
```
SimonMellerin, line 62:
```suggestion
        foreach (($config['storage']['filename_strategy'] ?? []) as $connectionName => $strategyId) {
```

SimonMellerin (changes requested):
Maybe you can add some unit tests :)

src/Storage/FilenameStrategyInterface.php
```diff
@@ -0,0 +1,27 @@
...
+    ): string;
+
+    /**
+     * If your strategy doesn't store backups in the configured root directory
+     * then return here a root path in which backups will be lookup by the
+     * storage manager when the restore command is run.
+     *
+     * Most implementation should just return "null" here.
+     */
+    public function getRootDir(): ?string;
```
SimonMellerin, line 26: This method name is confusing but I don't know how to make it more explicit

pounard: At least the name is consistent with the parameter name, I have better name myself.

SimonMellerin (approved):

src/Storage/FilenameStrategyInterface.php
```diff
@@ -0,0 +1,27 @@
...
+
+namespace MakinaCorpus\DbToolsBundle\Storage;
+
+interface FilenameStrategyInterface
+{
+    /**
+     * Generate dump file name.
+     */
+    public function generateFilename(
+        string $storageRootDir,
```
Lonnytunes, line 13: Is the storage root directory really necessary? Could we not just expect a filename relative to this root from implementations?

PR #82, opened 2024-02-12
no-issue - fix Anonymize cmd 

SimonMellerin:
backup and restore initial database in non interactive mode

### DIFF
--- a/src/Backupper/PgSQL/Backupper.php
+++ b/src/Backupper/PgSQL/Backupper.php
@@ -43,8 +43,9 @@ class Backupper extends AbstractBackupper
         $args[] = '-Z'; // compression level 0-9
         $args[] = '5';
         $args[] = '--lock-wait-timeout=120';
-        $args[] = '--exclude-table-data=' . \implode('|', $this->excludedTables);
-        $args[] = '--blobs';
+        if (\count($this->excludedTables)) {
+            $args[] = '--exclude-table-data=' . \implode('|', $this->excludedTables);
+        }
         $args[] = $dbParams['dbname'];
 
         if ($this->verbose) {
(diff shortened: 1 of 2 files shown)


src/Command/Anonymization/AnonymizeCommand.php
```diff
@@ -33,7 +33,7 @@ class AnonymizeCommand extends Command
 
     // Command behavior
     private bool $doAnonymizeCurrentDatabase = false;
-    private ?bool $doBackupAndRestoreInitial = null;
+    private ?bool $doBackupAndRestoreInitial = true;
```
pounard, line 36: Why is `null` allowed via `?bool` if there is a default value ?

src/Command/Anonymization/AnonymizeCommand.php
```diff
@@ -173,7 +173,7 @@ protected function interact(InputInterface $input, OutputInterface $output): voi
             return;
         }
 
-        if (null === $this->doBackupAndRestoreInitial) {
+        if (false !== $this->doBackupAndRestoreInitial) {
```
pounard, line 176: Why not simply `if ($this->doBackupAndRestoreInitial)` ? Is `null` still allowed (cf. my comment upper) ?

src/Backupper/PgSQL/Backupper.php
```diff
@@ -43,8 +43,9 @@ public function startBackup(): self
         $args[] = '-Z'; // compression level 0-9
         $args[] = '5';
         $args[] = '--lock-wait-timeout=120';
-        $args[] = '--exclude-table-data=' . \implode('|', $this->excludedTables);
-        $args[] = '--blobs';
+        if (\count($this->excludedTables)) {
```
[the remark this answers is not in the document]
SimonMellerin: yes but I find it more explicit ?

pounard (approved):

PR #83, opened 2024-02-12
feat: ignore MappedSuperclass entities

DjLeChuck:
This kind of entity is used to be extended by others, so they must be ignored otherwise anonymizations try to be carried out on them which does not work.

For example:

* MappedSuperclass Person
* Entity Employee extends Person
* Entity Correspondant extends Person

The anonymization try to process thre tables: person, employee and correspondant whereas only two exists: employee and correspondant


SimonMellerin:
Thanks for this, we missed that one!

SimonMellerin (approved):

PR #84, opened 2024-02-12
fix: wrong condition on cancel confirmation

DjLeChuck:
The question `You are about to erase your local database. Are you sure you want to continue?` wants us to answer `yes` in order to continue; however the condition is not inverted so we must answer `no` to continue.


SimonMellerin:
Thanks !

SimonMellerin (approved):

PR #88, opened 2024-02-13
No issue - Test the bundle configuration

Lonnytunes:


tests/Resources/config/packages/db_tools_alt2.yaml
```diff
@@ -0,0 +1,28 @@
+db_tools:
+    storage:
+        root_dir: '%kernel.project_dir%/var/backup'
+        filename_strategy:
+            connection_two: app.db_tools.custom_filename_strategy
+    backup_expiration_age: '6 months ago'
```
pounard, line 6: This value should go into `storage.expires_after` maybe ?

pounard:
Very nice PR,  we need this.

tests/Resources/config/packages/db_tools_alt2.yaml
Lonnytunes: We could move it in the storage section, indeed, but I would prefer to keep its current name.

However, it's not essential for now. If we decide to make this change, I propose to plan it for a future version.

pounard: Yeah, just leave it as-is, but it's a direct storage manager parameter in the end, I think it would make sense to move it into the storage section.

SimonMellerin (changes requested):
Some of the fixes you made in `config/packages/db_tools.yaml` should probably also been made in `docs/content/configuration.md` and maybe in `docs/content/backup_restore.md`

Lonnytunes:
> Some of the fixes you made in `config/packages/db_tools.yaml` should probably also been made in `docs/content/configuration.md` and maybe in `docs/content/backup_restore.md`

There was indeed a little fix to do in  `docs/content/configuration.md`  :+1: That's done.

Issue #95, opened 2024-02-20
Hash function anonymizer, add a salt to hashed values


pounard:
Hashing is not encrypting, and hashes can be reversed as simply as googling them.

We need to use a global salt:

 - salting values will prevent reversing,
 - global salt will keep hashes reproducible across the whole database.

Basically, we need to:

 - create a hash at the beginning of the procedure, and inject it to all anonymizers,
 - replace SQL code such as `MD5(foo.bar)` using `MD5(foo.bar || ?)` where the `?` placeholder will then be the salt.

Easy, let's do this !

PR #96, opened 2024-02-21
issue #95 - add salt to md5 and email anonymizers

pounard:


docs/content/anonymization/core-anonymizers.md
```diff
@@ -254,6 +262,17 @@ customer:
 ```
 :::
 
+:::tip
+Hashing a string is not anonymizing it because hash functions have a reproducible output,
+this in order to avoid decrypting data using rainbow tables, per default a salt will be
+added to string values prior to hasing.
```
Lonnytunes, line 268:
```suggestion
Hashing a string is not anonymizing it because hash functions have a reproducible output.
In order to avoid decrypting data using rainbow tables, a salt will be added by default to
string values prior to hashing.
```

docs/content/anonymization/core-anonymizers.md
```diff
@@ -81,6 +81,14 @@ customer:
 ```
 :::
 
+:::tip
+Username value is salted prior to be hashed using md5 in order to prevent reverse hashing
+with rainbow tables, but salt is global across the same anonymizer run, which keeps a single
+value when the same email is anonymized more than once.
```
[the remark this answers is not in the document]
pounard: Actually I changed "username" to "email" (will push that soon) it's more correct.

docs/content/backup_restore.md
```diff
@@ -32,10 +32,10 @@ console db-tools:backup --connection other_connection_name
 
 ### Excluded tables
 
-You may have configured [tables to be excluded in the bundle configuration](./configuration#excluded-tables).
+You may have configured [tables to be exclude in the bundle configuration](./configuration#excluded-tables).
```
Lonnytunes, line 35:
```suggestion
You may have configured [tables to exclude in the bundle configuration](./configuration#excluded-tables).
```

pounard: Not my PR !

pounard: Sorry, I forgot to rebase actual main branch. Fixed.

src/Anonymization/Anonymizer/Core/EmailAnonymizer.php
```diff
@@ -14,6 +14,7 @@
     pack: 'core',
     description: <<<TXT
     Anonymize email addresses. You can choose a domain and a tld with option 'domain'.
+    Values are salted to avoid reverse hashing with option 'use_salt' (default true).
```
Lonnytunes, line 17:
```suggestion
    Values are salted to avoid reverse hashing with option 'use_salt' (default: true).
```

src/Anonymization/Anonymizer/Core/Md5Anonymizer.php
```diff
@@ -11,7 +11,11 @@
 #[AsAnonymizer(
     name: 'md5',
     pack: 'core',
-    description: 'Anonymize a column by hashing its value.'
+    description: <<<TXT
+    Anonymize a column by hashing its value.
+    Options are 'use_salt' (default true).
+    Using a salt prevents reverse hash of values using rainbow tables.
```
Lonnytunes, line 17: I think "... prevents reversing (the) hash of ..." or "... prevents reverse hashing of ..." would be a more correct wording.

src/Anonymization/Anonymizator.php
```diff
@@ -52,12 +58,17 @@ protected function createAnonymizer(AnonymizerConfig $config): AbstractAnonymize
...
             $config->table,
             $config->targetName,
             $this->connection,
             $config->options,
         );
+        \assert($ret instanceof AbstractAnonymizer);
+
+        $ret->setSalt($this->getSalt());
+
+        return $ret;
```
SimonMellerin, line 71: Every `Anonymizer` has to implement `AbstractAnonymizer`, so we could just do

```suggestion
        return new $className(
            $config->table,
            $config->targetName,
            $this->connection,
            $config->options,
            $this->getSalt(),
        );
        return $ret;
```

no ?

src/Anonymization/Anonymizer/AbstractAnonymizer.php
```diff
@@ -27,8 +27,14 @@ final public function __construct(
         protected string $columnName,
         protected Connection $connection,
         protected Options $options,
+        private ?string $salt = null,
```
SimonMellerin, line 30: Why not 
```suggestion
        private string $salt,
```
?

And put the `::generateRandomSalt` method on the `Anonymizator` class ?

docs/content/anonymization/core-anonymizers.md
```diff
@@ -254,6 +262,17 @@ customer:
 ```
 :::
 
+:::tip
```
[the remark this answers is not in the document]
pounard: As you wish, both are fine because it also document how to deactivate, but let's go for info.

src/Anonymization/Anonymizator.php
pounard: Yes, indeed, I'll change that.

src/Anonymization/Anonymizer/AbstractAnonymizer.php
pounard: I'm not sure, anonymizers could be services one day, and I don't see why we should then force the salt to be in the constructor ?

SimonMellerin: makes sens, then maybe we should push it throw options ?

pounard: If the correct word was "thought" then yes, it'd also make sense, as you prefer I have no strong opinion. Maybe we'd need some kind of common validation callback for options, that would validate all common options ?

pounard:
Ah ! Interesting, since I add the string concat, some RDBMS seem to ignore that NULL values are NULL. I need to add a `$this->getSetIfNotNullExpression()` call.

src/Anonymization/Anonymizer/AbstractAnonymizer.php
SimonMellerin: One of the goal of this class was to not have to update the AbstractAnonymizer constructor signature in the future and as this property will only affect a few anonymizers, I prefer that it not pass the options.

SimonMellerin: And yes, it could be a good idea to had a method `AbstractAnonymizer::validateOptions(Options $options)`  to let each anonymizer validate the options you give to it

src/Anonymization/Anonymizer/Core/Md5Anonymizer.php
```diff
@@ -21,10 +25,21 @@ class Md5Anonymizer extends AbstractAnonymizer
...
+        if ($this->options->get('use_salt', true)) {
+            $columnExpr = $expr->concat($columnExpr, $expr->value($this->getSalt()));
+
+            $update->set(
+                $this->columnName,
+                // Work around some RDBMS not seeing the NULL value anymore
+                // once we added the string concat.
+                $this->getSetIfNotNullExpression(
+                    $expr->md5($columnExpr)
+                ),
```
SimonMellerin, line 39: why is this needed here and not on the EmailAnonymizer ? :thinking:

src/Anonymization/Anonymizer/AbstractAnonymizer.php
SimonMellerin: yep

src/Anonymization/Anonymizer/Core/Md5Anonymizer.php
pounard: It is already in email anonymizer if I remember correctly.

pounard: Yes, I can confirm that.

SimonMellerin (approved):

Issue #79, opened 2024-02-11
Allow to specify "extra arguments" to a backupper service


jdecool:
When dumping large databases, I generally add specify some arguments.

For example, with `mysqldump` I want to have the possibility to make a dump with `--single-transaction` and `--skip-lock-tables` arguments to avoid disturbing my website availability.

SimonMellerin:
That could indeed be, a good addition.

May be you could add:

- [ ] Configurations in `config/packages/db_tools.yaml` like `backupper_arguments` to completely overwrite default ones
- [ ] An option for the `db-tools:backup` to specify extra arguments on the fly

We will try do that soon because it sounds like it will be a common need, thanks for the suggestion.

pounard:
Implement both is probably the way to go: default arguments in YAML is a nice to have, configure it once, run all the time. But the idea of being able to pass arbitrary options when manually running the script is a good idea as well.

SimonMellerin:
And, this should obviously be done for backupper *and* restorer

SimonMellerin:
This has been implemented by @Lonnytunes (see #99 )

It is already on the `main` branch and will be available in the next release (`1.1.x` expected for the next week)

jdecool:
Awesome !

Thanks.

PR #98, opened 2024-02-21
Issue #79 - Complete or replace default options

Lonnytunes:
Add the `--replace-defaults` option to backup and restoration commands to control the behavior of the `--extra-options` option.

### DIFF
--- a/docs/content/backup_restore.md
+++ b/docs/content/backup_restore.md
@@ -60,32 +60,25 @@ Note that using this option, backup files will never be cleaned up.
 
 ### Extra options
 
-If you temporarily want to provide your own options to the backup binary,
-use the `--extra-options` (`-o`) in your command:
+If you need to occasionally provide some custom options to the backup binary,
+use the `--extra-options` (`-o`) option in your command:
 
 ```sh
 console db-tools:backup --extra-options "--opt1 val1 --opt2 val2 --flag"
 ```
 
-If you always need to use the same custom options when backing up, you can
-define them as default options in the bundle configuration file:
+Unless you specify the `--ignore-default-options` option, the custom options
+will be added to the [default options](./configuration#default-binary-options).
 
-```yaml
-# config/packages/db_tools.yaml
-db_tools:
-    # ...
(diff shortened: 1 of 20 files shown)


docs/content/backup_restore.md
```diff
@@ -60,31 +60,25 @@ Note that using this option, backup files will never be cleaned up.
...
+use the `--extra-options` (`-o`) option in your command:
 
 ```sh
 console db-tools:backup --extra-options "--opt1 val1 --opt2 val2 --flag"
 ```
 
-If you always need to use the same custom options when backing up, you can
-define them as default options in the bundle configuration file:
+By default, those custom options will be added to the [default options](./configuration#default-binary-options)
+defined in the bundle configuration file. Use `--replace-defaults` (`r`) if you
```
SimonMellerin, line 71: may be `--override-options`

I don't know what is better :shrug:

docs/content/backup_restore.md
```diff
@@ -60,31 +60,25 @@ Note that using this option, backup files will never be cleaned up.
...
-See the section about the [bundle configuration](./configuration#default-binary-options)
-for more information.
-
 :::warning
-The `--extra-options` option overrides default options defined in the
-configuration file as well as those from the bundle itself. (No merge
-will be performed.)
+Whether the `--replace-defaults` option is used or not and whether default
+options are defined in the configuration file or not, `--extra-options` option
+value will always override default options from the bundle.
```
SimonMellerin, line 81: I don't understand this sentence? are you sure about it?

SimonMellerin: Hum ok I think I see what you mean. I understand why this operate like this (because I know the code behind it), but It's not easy to understand.. :/

SimonMellerin: Maybe Backupper and Restorer could have 2 properties: $defaultOptions and $extraOptions ?

Lonnytunes:
After our discussion IRL and the changes I've made on the code, I prefer to close this PR and create a new one.

PR #99, opened 2024-02-22
Issue #79 - Revision of default and extra options management

Lonnytunes:


src/Backupper/AbstractBackupper.php
```diff
@@ -16,13 +17,18 @@ abstract class AbstractBackupper implements \IteratorAggregate
 {
     protected ?string $destination = null;
     protected array $excludedTables = [];
+    protected string $defaultOptions = '';
```
SimonMellerin, line 20: As you initialize the property in constructor, I don't know if this is not better:
```suggestion
    protected string $defaultOptions;
```

SimonMellerin: (maybe we don't care)

src/Backupper/AbstractBackupper.php
```diff
@@ -84,6 +90,18 @@ public function getExtraOptions(?string $options): ?string
...
     }
 
+    public function ignoreDefaultOptions(bool $switch = true): self
+    {
+        $this->ignoreDefaultOptions = $switch;
+
+        return $this;
+    }
+
+    public function areDefaultOptionsIgnored(): bool
```
SimonMellerin, line 100: is this method necessary?

src/Backupper/AbstractBackupper.php
```diff
@@ -96,6 +114,28 @@ public function isVerbose(): bool
         return $this->verbose;
     }
 
+    /**
+     * Provide the built-in default options that will be used if none is given
+     * through the dedicated constructor argument.
+     */
+    protected function getBuiltinDefaultOptions(): string
+    {
+        return '';
```
SimonMellerin, line 123: I don't know if this method should not be abstract to force people to think about it when they add a new backupper ?

src/Backupper/AbstractBackupper.php
```diff
@@ -96,6 +114,28 @@ public function isVerbose(): bool
...
+    }
+
+    /**
+     * Add default (if not ignored) and extra options to the given command line.
+     */
+    protected function addCustomOptions(CommandLine $command): void
```
SimonMellerin, line 129: may be this could be `final`

src/Backupper/SQLite/Backupper.php
```diff
@@ -23,17 +23,17 @@ public function startBackup(): self
...
-                $tablesToBackup,
-                $this->binary,
-                $this->extraOptions ?: '-bail',
-                $dbParams['path'],
-                \addcslashes($this->destination, '\\"')
-            ),
-            false
-        );
+        // The CommandLine instance below will generate something like:
+        // echo 'BEGIN IMMEDIATE;\n.dump table1 table2 ...' | 'sqlite3' -bail > '/path/to/backup.sql'
```
SimonMellerin, line 27: :+1:

src/Backupper/SQLite/Backupper.php
```diff
@@ -82,4 +82,10 @@ private function getTablesToBackup(): array
 
         return $tables;
     }
+
+    #[\Override]
+    protected function getBuiltinDefaultOptions(): string
+    {
+        return '-bail';
```
SimonMellerin, line 89: I don't even know what this is ^^

src/Backupper/AbstractBackupper.php
SimonMellerin: yes indeed!

src/Backupper/SQLite/Backupper.php
Lonnytunes: The SQLite help gives this description: "stop after hitting an error".

src/Backupper/AbstractBackupper.php
Lonnytunes: Seems relevant. I add it.

src/Backupper/AbstractBackupper.php
SimonMellerin: Forget this comment, we'll see that later

src/Backupper/AbstractBackupper.php
Lonnytunes: I chose to avoid having to implement it for "nothing", but "force people to think about it when they add a new backupper" seems not absurd :wink: I can change that if you prefer that we do it this way.

src/Backupper/AbstractBackupper.php
Lonnytunes: I always ask myself this same question every time I meet this case. I don't know what is the best practice.

src/Backupper/AbstractBackupper.php
SimonMellerin: I have no strong opinion, if was just a suggestion, as you like!

SimonMellerin (approved):

src/Backupper/AbstractBackupper.php
Lonnytunes: I keep the current version. This can be easily changed later.

PR #100, opened 2024-02-23
No issue - Refactor code related to backuppers and restorers

Lonnytunes:


Lonnytunes:
@pounard @SimonMellerin

I created the directory `src/Utility` in a previous PR for the `CommandLine` class.

I added `AbstractProcessTrait` (to rename?) and `ChainLogger` elements into this directory for now.

But I think `CommandLine` and `AbstractProcessTrait` could be in an `src/Process` directory, and `ChainLogger` in an `src/Log`.

Let me know your opinions! :slightly_smiling_face:

src/Backupper/AbstractBackupper.php
```diff
@@ -43,9 +46,10 @@ public function checkBinary(): string
     {
         $process = new Process([$this->binary, '--version']);
         $process->run();
+
         if (!$process->isSuccessful()) {
-            throw new \InvalidArgumentException(\sprintf(
-                "Error while trying to process '%s', check configuration for binary '%s",
+            throw new \LogicException(\sprintf(
+                "Error while running \"%s\" command, check configuration for binary '%s'.",
```
pounard, line 52: That's kind of schizophrenic, either enclose with `"` and use `'` or the other way around, but mixing the two is weird.

src/Backupper/BackupperFactory.php
```diff
@@ -22,6 +24,7 @@ public function __construct(
         private ManagerRegistry $doctrineRegistry,
         private array $backupperBinaries,
         private array $backupperOptions = [],
+        private ?LoggerInterface $logger = null,
```
pounard, line 27: You should make it non-nullable and create a null instance here, no ? That's not mandatory but I dislike dangling nullables for such services. I don't know if this would make any sense to instrument the factories themselves ?

src/Command/Anonymization/AnonymizeCommand.php
```diff
@@ -355,27 +332,16 @@ private function doBackupAnonymizedDatabase(): void
...
-        foreach ($backupper as $data) {
-            if ($this->io->isVerbose()) {
-                $this->io->text($data);
-            }
-        }
-
-        $backupper->checkSuccessful();
         if ($this->io->isVerbose()) {
-            $this->io->text($backupper->getOutput());
+            $backupper->setVerbose(true)->addLogger(new ConsoleLogger($this->io));
```
pounard, line 336: Don't use a setVerbose() option on backupers, simply always log, but using level. The logger configuration will or will not display items depending on their level. That's useless logic IMHO.

src/Command/BackupCommand.php
```diff
@@ -112,23 +114,18 @@ private function doBackup(): string
...
-        $this->backupper
+        $this
+            ->backupper
             ->setDestination($filename)
-            ->setVerbose($this->io->isVerbose())
             ->setExcludedTables($this->excludedTables[$this->connectionName] ?? [])
             ->setExtraOptions($this->extraOptions)
             ->ignoreDefaultOptions($this->ignoreDefaultOptions)
-            ->startBackup()
+            ->setVerbose($this->io->isVerbose())
```
pounard, line 123: Same, I'm not sure verbosity is something the backuper should know, simply let it log everything, using logger levels.

src/Backupper/BackupperFactory.php
Lonnytunes: I chose to make it "nullable" to avoid adding a useless `NullLogger` into the backupper's `ChainLogger`. I usually don't like having to deal with `null` values as well, but in this case of a short factory I consider that this is acceptable.

src/Command/Anonymization/AnonymizeCommand.php
Lonnytunes: I just kept the initial logic as it was, but it's actually no longer relevant now that we use loggers. I will change that.

src/Backupper/AbstractBackupper.php
Lonnytunes: I changed this because the command will probably contain a lot `'` on Linux. I can also change it for the second placeholder if you prefer.

src/Backupper/BackupperFactory.php
pounard: OK ! That makes sense.

src/Command/BackupCommand.php
Lonnytunes: I close this discussion which is a duplicate.

src/Command/Anonymization/AnonymizeCommand.php
Lonnytunes: In fact, `setVerbose()` is not initially used to control the logging verbosity of the backupper itself, it is mainly to tell the backupper to call the binary in verbose mode. I will keep it, but I will add the logger independently of the verbosity level, `ConsoleLogger` handles verbosity by itself.

pounard:
@Lonnytunes 

> But I think CommandLine and AbstractProcessTrait could be in an src/Process directory, and ChainLogger in an src/Log.

I have no strong opinion about this. For the first part, the `Process` namespace is a good idea, for the other one, I think that the `Utility` namespace must be removed, and the `ChainLogger` moved into the `Helper` namespace, which is basically the same.

Let's keep only one "fourzitou" namespace.

Lonnytunes:
Thanks for you answer @pounard 

I will move `ChainLogger` from the `Utility` namespace to the `Helper` one. It's actually a kind of duplicate.

But I prefer to keep the idea of the `Process` namespace at the root for now.

Are you agree to rename `AbstractProcessTrait` to `ProcessTrait`?

pounard:
> Are you agree to rename AbstractProcessTrait to ProcessTrait?

Yes, it's either abstract, either a trait, but not both.

src/Backupper/AbstractBackupper.php
```diff
@@ -43,9 +46,10 @@ public function checkBinary(): string
     {
         $process = new Process([$this->binary, '--version']);
         $process->run();
+
         if (!$process->isSuccessful()) {
-            throw new \InvalidArgumentException(\sprintf(
-                "Error while trying to process '%s', check configuration for binary '%s",
+            throw new \LogicException(\sprintf(
```
pounard, line 51: Maybe a `\RuntimeException` here ?

src/Command/Anonymization/AnonymizeCommand.php
```diff
@@ -247,31 +248,20 @@ private function doBackupInitialDatabase(): void
...
         $backupper = $this->backupperFactory->create($this->connectionName);
-
-        $this->initialDatabaseBackupFilename = $this->storage->generateFilename($this->connectionName, $backupper->getExtension());
+        $this->initialBackupFilename = $this->storage->generateFilename($this->connectionName, $backupper->getExtension());
 
         $backupper
-            ->setDestination($this->initialDatabaseBackupFilename)
+            ->setDestination($this->initialBackupFilename)
+            ->addLogger(new ConsoleLogger($this->io))
             ->setVerbose($this->io->isVerbose())
```
pounard, line 256: You should remove the setVerbose stuff, I really do think it's not relevant anymore.

src/Restorer/AbstractRestorer.php
```diff
@@ -34,9 +37,10 @@ public function checkBinary(): string
     {
         $process = new Process([$this->binary, '--version']);
         $process->run();
+
         if (!$process->isSuccessful()) {
-            throw new \InvalidArgumentException(\sprintf(
-                "Error while trying to process '%s', check configuration for binary '%s",
+            throw new \LogicException(\sprintf(
```
pounard, line 42: I would use a `\RuntimeException` as well here. `\LogicException` kind of means that the user did something wrong, that may not be the case here.

src/Restorer/AbstractRestorer.php
```diff
@@ -93,6 +97,12 @@ public function isVerbose(): bool
         return $this->verbose;
     }
 
+    #[\Override]
+    protected function beforeProcess(): void
+    {
+        $this->process->setTimeout(1800);
```
pounard, line 103: I'm still concerned about this timeout, it should be configurable.

src/Command/Anonymization/AnonymizeCommand.php
Lonnytunes: As I said previously in the related discussion:

> In fact, setVerbose() is not initially used to control the logging verbosity of the backupper itself, it is mainly to tell the backupper to call the binary in verbose mode. I will keep it, but I will add the logger independently of the verbosity level, ConsoleLogger handles verbosity by itself.

src/Restorer/AbstractRestorer.php
Lonnytunes: I agree, but this is the default timeout. Make the timeout configurable is another subject.

src/Restorer/AbstractRestorer.php
Lonnytunes: I hesitated. Ok to change that.

src/Restorer/AbstractRestorer.php
Lonnytunes: https://github.com/makinacorpus/DbToolsBundle/issues/102

pounard (approved):

Issue #85, opened 2024-02-13
IntegerAnonymizer - add possibility to anonymize value by adding noise


SimonMellerin:
Add options to IntegerAnonymizer to be able to anonymize int by adding noise: a delta or a percent.

```php
    #[ORM\Column]
    // This will add to the value a random value between -10 and 10
    #[Anonymize(type: 'integer', options: ['delta' => '10'])]
    // This will add to the value a random value between -10 and 10 and will keep an absolute value
    #[Anonymize(type: 'integer', options: ['delta' => '10', 'unsigned' => true])]
    // This will multiply the value by a random value between 0.9 and 1.1
    #[Anonymize(type: 'integer', options: ['percent' => 10])]
    private ?int $age = null;
```

SimonMellerin:
done with #110

PR #110, opened 2024-03-08
#85 - IntegerAnonymizer - add possibility to anonymize value by adding noise

SimonMellerin:


src/Anonymization/Anonymizer/Core/IntegerAnonymizer.php
```diff
@@ -23,18 +26,67 @@ class IntegerAnonymizer extends AbstractAnonymizer
...
+        if ($this->options->has('min') && $this->options->has('max')) {
+            $this->anonymizeWithMinAndMax(
+                $update,
+                (int) $this->options->get('min'),
+                (int) $this->options->get('max')
+            );
+        } elseif ($this->options->has('delta')) {
+            $this->anonymizeWithDelta($update, (int) $this->options->get('delta'));
+        } elseif ($this->options->has('percent')) {
+            $this->anonymizeWithPercent($update, (int) $this->options->get('percent'));
```
Lonnytunes, line 39: Shouldn't we ensure that `delta` and `percent` options contain a positive integer ?

Or maybe we could just add `$delta = \abs($delta)` before using it in `anonymizeWithDelta()`? The same thing for the percent version.

src/Anonymization/Anonymizer/Core/IntegerAnonymizer.php
```diff
@@ -23,18 +26,67 @@ class IntegerAnonymizer extends AbstractAnonymizer
...
+
+    private function anonymizeWithPercent(Update $update, int $percent): void
+    {
+        $expr = $update->expression();
+
+        $update->set(
+            $this->columnName,
+            $expr->cast(
+                $expr->raw(
+                    '? * (?) / 100',
```
Lonnytunes, line 79: I wonder about the parentheses. Are they really useful?

SimonMellerin: yep, it does not work without them

src/Anonymization/Anonymizer/Core/IntegerAnonymizer.php
SimonMellerin: I think we should trigger an exception if percent or delta are negative. Putting such values for those options will be (most of the time) the result of a mistake or a misunderstood on how this anonymizer works.

src/Anonymization/Anonymizer/Core/IntegerAnonymizer.php
Lonnytunes: Quite strange :thinking:

SimonMellerin: Well i guess it depends of what the `$this->getRandomIntExpression` returns (in SQL). 

It only failed for Mysql I think.

SimonMellerin: No I think it was SQLite :

```php

        return $this->formatRaw(
            new Raw(
                'abs(random()) % (? - ?) + ?',
                [$max, $min, $min]
            ),
            $context,
        );
```
(from the query builder source code)

Lonnytunes: :+1:

Lonnytunes: Finally, shouldn't we fix this method from the `SqliteWriter` instead?

SimonMellerin: see https://github.com/makinacorpus/php-query-builder/issues/6

src/Anonymization/Anonymizer/Core/IntegerAnonymizer.php
Lonnytunes: I agree with this solution :ok_hand:

src/Anonymization/Anonymizer/Core/IntegerAnonymizer.php
Lonnytunes: > I agree but as it comes from a dependency, I suggest we keep it like this for now

Very true :sweat_smile:

Lonnytunes (approved):

Issue #97, opened 2024-02-21
Anonymizers options improvements


pounard:
As stated in https://github.com/makinacorpus/DbToolsBundle/pull/96#discussion_r1497552466 there might some room for improvements in options handling:

 - global options validation in `AbstractAnonymizer`, we don't have any global options right now, but we might have later,
 - custom validation, ie. `AbstractAnonymizer::validateOptions(Options $options)` for each anonymizer to provide validation for its own options,
 - ~~question remains, should the seed/salt be injected in options ?~~ done in https://github.com/makinacorpus/DbToolsBundle/pull/96

SimonMellerin:
This validation could also be used by the `db-tools:anonymization:dump-config` command.

pounard:
Exactly what I was thinking about ! We probably should not block in case of invalid options for production operation.

pounard:
Also, I'd like to add helpers on the `Option` class, such as:

 - `getInt(string $name, int $default, bool $required = false): ?int;`
 - `getFloat(string $name, float $default, bool $required = false): ?float;`
 - `getDate(string $name, \DateTimeImmutable $default, bool $required = false): ?\DateTimeImmutable;`
 - `getInterval(string $name, \DateInterval $default, bool $required = false): ?\DateInterval;`
 - etc...

And add new helper along the way for anonymizers to use.

@SimonMellerin @Lonnytunes what is your opinion about this?

SimonMellerin:
I'm not sure about this, I prefer to keep things simple for now and not over-anticipate needs

pounard:
It would validates the user input, and I already copy/pasted code (for example the parseDate()).

SimonMellerin:
This has been implemented, see #128 ans #131

It is already on the `main` branch and will be available in the next release (`1.1.x` expected for the next week)

PR #113, opened 2024-03-08
#86 - FloatAnonymizer - add possibility to anonymize value by adding noise

SimonMellerin:
To merge and review once #108 will be merged

### DIFF
--- a/src/Anonymization/Anonymizer/Core/FloatAnonymizer.php
+++ b/src/Anonymization/Anonymizer/Core/FloatAnonymizer.php
@@ -12,8 +12,12 @@ use MakinaCorpus\QueryBuilder\Query\Update;
     name: 'float',
     pack: 'core',
     description: <<<TXT
-    Anonymize with a random float between two bounds.
-    Options are 'min' , 'max' and 'precision' (default 2).
+    Anonymize float by:
+        - randomly choosing an integer in a range delimited
+          by 'min' and 'max' options
+        - altering the initial value by adding it a random value picked
+          in a range computed from the 'delta' or 'percent' options
+    You may also specify a 'precision' (default 2).
     TXT
 )]
 class FloatAnonymizer extends AbstractAnonymizer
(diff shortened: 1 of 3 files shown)


src/Anonymization/Anonymizer/Core/FloatAnonymizer.php
```diff
@@ -23,27 +27,109 @@ class FloatAnonymizer extends AbstractAnonymizer
...
+        if ($precision <= 0) {
+            throw new \InvalidArgumentException("'precision' should be greater than 0.");
         }
+        $precision = 10 ** $precision;
+
+        if ($this->options->has('min') && $this->options->has('max')) {
+            $min = (float) $this->options->get('min');
+            $max = (float) $this->options->get('max');
+            if ($min >= $max) {
+                throw new \InvalidArgumentException("'max' should be greater than 'min'.");
```
pounard, line 41: Honestly, I would simply inverse both instead of raising an exception, and trigger a warning instead ?

src/Anonymization/Anonymizer/Core/FloatAnonymizer.php
```diff
@@ -23,27 +27,109 @@ class FloatAnonymizer extends AbstractAnonymizer
...
+            }
+
+            $this->anonymizeWithMinAndMax($update, $precision, $min, $max);
+        } elseif ($this->options->has('delta')) {
+            $delta = (float) $this->options->get('delta');
```
pounard, line 46: You maybe should check if `'percent'` is present too, and raise an exception if both are specified.

src/Anonymization/Anonymizer/Core/FloatAnonymizer.php
SimonMellerin: I never know what to do in such case.

In one hand, in most cases, you probably right, just inverse both could be a good solution. 
But what if it's a typo and `max` has a typo for example 100 instead of 1000 ? The error will pass silently :/

At least here, we warn :)

src/Anonymization/Anonymizer/Core/FloatAnonymizer.php
SimonMellerin: yep, what I'm gonna say is not coherent with my last comment but in fact, I thought to refin this with #97

src/Anonymization/Anonymizer/Core/FloatAnonymizer.php
pounard: I let you decide, whatever you choose it's OK for me. Triggering a warning in a logger is a good idea (we should inject a logger into anonymizers as well I guess).

SimonMellerin: Anonymizers are not services, it won't be so straight forward to inject logger in it. Let's keep it like this for now.

pounard (approved):

Issue #114, opened 2024-03-09
Add a NullAnonymizer


SimonMellerin:
It will just replaces data with 'NULL'

PR #116, opened 2024-03-11
#114 - Add NullAnonymizer

SimonMellerin:


src/Anonymization/Anonymizer/Core/NullAnonymizer.php
```diff
@@ -0,0 +1,29 @@
...
+{
+    /**
+     * @inheritdoc
+     */
+    public function anonymize(Update $update): void
+    {
+        $expr = $update->expression();
+        $update->set(
+            $this->columnName,
+            $expr->raw('NULL'),
```
pounard, line 26: Just write `null` here, no need to use an expression instance.

docs/content/anonymization/core-anonymizers.md
```diff
@@ -265,6 +265,41 @@ customer:
 ```
 :::
 
+## NullAnonymizer
+
+Set all value to `NULL`.
```
Lonnytunes, line 270:
```suggestion
Set all values to `NULL`.
```

src/Anonymization/Anonymizer/Core/NullAnonymizer.php
SimonMellerin: it does not work with just 'null'

SimonMellerin: I updated with `$expr->null()`

src/Anonymization/Anonymizer/Core/NullAnonymizer.php
```diff
@@ -0,0 +1,29 @@
...
+{
+    /**
+     * @inheritdoc
+     */
+    public function anonymize(Update $update): void
+    {
+        $expr = $update->expression();
+        $update->set(
+            $this->columnName,
+            $expr->null(),
```
pounard, line 26: You can simply pass `null` here, instead of using an expression.

SimonMellerin: no, it does not work with just `null`, I tested

Lonnytunes (approved):

Issue #115, opened 2024-03-09
Add a ConstantAnonymizer


SimonMellerin:
Replace all value by a single one.

Option is 'value', default value for this option is 'xxxxxx'

PR #119, opened 2024-03-12
#115 - Add ConstantAnonymizer

SimonMellerin:
to review and merge after #116

### DIFF
--- a/docs/content/anonymization/core-anonymizers.md
+++ b/docs/content/anonymization/core-anonymizers.md
@@ -294,7 +294,53 @@ class Customer
 # config/anonymization.yaml
 
 customer:
-    sensible_content: null
+    sensible_content: 'null'
+
+#...
+```
+:::
+
+## ConstantAnonymizer
+
+Set all value to `xxxxxx` or a custom value  (with the `value` option).
+
+::: code-group
+```php [Attribute]
+namespace App\Entity;
+
+use Doctrine\ORM\Mapping as ORM;
+use MakinaCorpus\DbToolsBundle\Attribute\Anonymize;
+
+#[ORM\Entity()]
+#[ORM\Table(name: 'customer')]
+class Customer
+{
+    // ...
+
+    #[ORM\Column]
+    #[Anonymize(type: 'constant')] // [!code ++]
+    private ?string $sensibleContent = null;
+
+    #[ORM\Column]
+    // You can specify a custom value with the `value` option
+    #[Anonymize(type: 'constant', options: ['value' => '_______'])] // [!code ++]
+    private ?string $sensibleContent = null;
+
+    // ...
+}
+```
+
+```yml [YAML]
+# config/anonymization.yaml
+
+customer:
+    sensible_content: constant
+
+customer:
+    # You can specify a custom value with the `value` option
+    sensible_content:
+        type: constant
+        options: {value: '_______'}
 
 #...
 ```
--- a/docs/content/anonymization/core-anonymizers.md
+++ b/docs/content/anonymization/core-anonymizers.md
@@ -302,7 +302,11 @@ customer:
 
 ## ConstantAnonymizer
 
-Set all value to `xxxxxx` or a custom value  (with the `value` option).
+Set all value to a constant value.
+Options are:
+* `value`: the value you want to use to fill the column
+* `type`: a SQL type for the given value
+  (default value is `text`)
 
 ::: code-group
 ```php [Attribute]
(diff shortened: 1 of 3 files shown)


docs/content/anonymization/core-anonymizers.md
```diff
@@ -265,6 +265,87 @@ customer:
 ```
 :::
 
+## NullAnonymizer
+
+Set all value to `NULL`.
```
Lonnytunes, line 270:
```suggestion
Set all values to `NULL`.
```

src/Anonymization/Anonymizer/Core/ConstantAnonymizer.php
```diff
@@ -0,0 +1,32 @@
...
+use MakinaCorpus\DbToolsBundle\Anonymization\Anonymizer\AbstractAnonymizer;
+use MakinaCorpus\DbToolsBundle\Attribute\AsAnonymizer;
+use MakinaCorpus\QueryBuilder\Query\Update;
+
+#[AsAnonymizer(
+    name: 'constant',
+    pack: 'core',
+    description: 'Set all value to \'xxxxxx\' or a custom value (with the \'value\' option).'
+)]
+class ConstantAnonymizer extends AbstractAnonymizer
```
pounard, line 16: You probably should add a "type" option, which then will be propagated as a `new Value($value, $type)` (or `$expr->value($value, $type)`), doing so will allow the query-builder PHP to SQL value converter to do its work.

src/Anonymization/Anonymizer/Core/NullAnonymizer.php
```diff
@@ -0,0 +1,29 @@
...
+)]
+class NullAnonymizer extends AbstractAnonymizer
+{
+    /**
+     * @inheritdoc
+     */
+    public function anonymize(Update $update): void
+    {
+        $expr = $update->expression();
+        $update->set(
```
pounard, line 24: Blank line here ?

src/Anonymization/Anonymizer/Core/NullAnonymizer.php
```diff
@@ -0,0 +1,29 @@
...
+            $this->columnName,
+            $expr->raw('NULL'),
```
pounard, line 26: Here you can write `$expr->null()` or simply a PHP `null` value, which will then be transparently propagated as an SQL `NULL` value.

docs/content/anonymization/core-anonymizers.md
```diff
@@ -265,6 +265,87 @@ customer:
...
+customer:
+    sensible_content: 'null'
+
+#...
+```
+:::
+
+## ConstantAnonymizer
+
+Set all value to `xxxxxx` or a custom value  (with the `value` option).
```
Lonnytunes, line 305:
```suggestion
Set all values to `xxxxxx` or a custom value  (with the `value` option).
```
I wonder if this default value is really pertinent. What if we make developers always define a value?

SimonMellerin: Yep, it was more pertinent when this anonymizer name was `BlankingAnonymizer`, like a marker stroke. But with this name, it makes no sense, you are right. I'm gonna update this

src/Anonymization/Anonymizer/Core/ConstantAnonymizer.php
SimonMellerin: done

src/Anonymization/Anonymizer/Core/ConstantAnonymizer.php
```diff
@@ -0,0 +1,44 @@
...
+        if (!$this->options->has('value')) {
+            throw new \InvalidArgumentException("'value' option should be given.");
+        }
+
+        $expr = $update->expression();
+
+        $update->set(
+            $this->columnName,
+            $this->getSetIfNotNullExpression(
+                $expr->cast(
```
pounard, line 37: Not `->cast()`, but `->value()` here !

`cast()` will simply add the `CAST(? AS type)` SQL expression, whereas `value()` will additionally trigger the PHP to SQL value converter.

pounard: And it could be nice to link the value converter documentation in the related anonymizer documentation page: https://php-query-builder.readthedocs.io/en/stable/converter/converter.html#input-value-converter

Lonnytunes (approved):

PR #122, opened 2024-03-14
feat: apply rector rules for return types

shakaran:
This PR helps with future PHP 8+ releases and return types.

Applied rector rules:
        ReturnTypeFromReturnNewRector
        ReturnTypeFromReturnDirectArrayRector
        ReturnTypeFromStrictBoolReturnExprRector
        ReturnTypeFromStrictConstantReturnRector
        ReturnTypeFromStrictNativeCallRector

PR #126, opened 2024-03-14
feat: apply rector rule import names over codebase

shakaran:
Apply rule [import names](https://getrector.com/documentation/import-names) with rector over all codebase. Follows to #122


SimonMellerin:
What this Rector rule does is not in our coding standards. 

I close this one.

PR #128, opened 2024-03-15
#97 - Add helpers on Options class to ease type validation

SimonMellerin:


src/Anonymization/Anonymizer/Options.php
```diff
@@ -51,4 +55,75 @@ public function toDisplayString(): string
             \array_values($this->options),
         ));
     }
+
+    public function getString(string $name, string $default = null, bool $required = false): ?string
+    {
+        $value = $this->get($name, $default, $required);
+        if ($value === null || is_scalar($value) || $value instanceof \Stringable) {
```
pounard, line 62: Are you considering null as an empty string ? You should always start with checking if null and not required then return null.

src/Anonymization/Anonymizer/Options.php
```diff
@@ -51,4 +55,75 @@ public function toDisplayString(): string
...
+
+        throw new \InvalidArgumentException(\sprintf("Option '%s' value should be a float", $name));
+
+    }
+
+    public function getDate(string $name, \DateTimeImmutable $default = null, bool $required = false): ?\DateTimeImmutable
+    {
+        $value = $this->get($name, $default, $required);
+
+        if ($value instanceof \DateTimeInterface) {
```
pounard, line 96: Aren't our configuration options not always strings ? Anyway better be safe than sorry I'm not against keeping this check.

SimonMellerin: I try to anticipate that in future, one could configure anonymization with some PHP.

Also, we do not validate that the array given in `Options` class constructor is string only, so I added this check just in case :shrug:

src/Anonymization/Anonymizer/Options.php
SimonMellerin: I wonder if we should consider `null` as empty or not ?

(I updated tests in order to accept `null` values but not sure if it is a good choice..)

SimonMellerin: `null` handling added

pounard: It seems cleaner to me, nice.

pounard (approved):

PR #134, opened 2024-03-15
No issue - Inject excluded tables configuration into the backupper factory instead of the backup command

Lonnytunes:


src/Backupper/BackupperFactory.php
```diff
@@ -58,6 +96,9 @@ public function create(?string $connectionName = null): AbstractBackupper
 
         \assert($backupper instanceof AbstractBackupper);
 
+        if (isset($this->excludedTables[$connectionName])) {
```
SimonMellerin, line 99: Don't know why, but I prefer `\key_exists($connectionName, $this->excludedTables)`

Do you know if one is better than the other?

src/Backupper/AbstractBackupper.php
```diff
@@ -78,7 +78,18 @@ public function getDestination(): ?string
 
     public function setExcludedTables(array $excludedTables): self
     {
-        $this->excludedTables = $excludedTables;
+        foreach ($excludedTables as $table) {
```
SimonMellerin, line 81: I think you should reset `$this->excludedTables` here.

Or the method name should be `::addExcludedTables`. But, if we consider how you use this method in `BackupCommand`, I think it should definitively be a `setExcludedTable` with a reset.

```suggestion
        $this->excludedTables = [];
        foreach ($excludedTables as $table) {
```

SimonMellerin: > If given, overrides excluded tables parameter from bundle configuration.

From the command description for the `excluded-table` option

SimonMellerin (changes requested):

src/Backupper/AbstractBackupper.php
Lonnytunes: > I think you should reset $this->excludedTables here.

Indeed! I applied your suggestion :pray:

SimonMellerin (approved):

Issue #105, opened 2024-03-05
Support Doctrine Embeddables with AttributeLaader


SimonMellerin:
For now, when setting up anonymization with PHP Attributes on Doctrine Entities, embeddables are skipped.
(see #101)
 
Workaround  is to use [yaml config](Embeddables) for these use cases.

SimonMellerin:
done with #105, 

It will be available in the next major release.

PR #145, opened 2024-04-19
#105 - AttributeLoader - Handle Doctrine Embeddables

SimonMellerin:


pounard (approved):
Globally I'm OK with this MR, are you sure this works with both ORM ^2 and ^3.0 ?

PR #146, opened 2024-04-20
fix: typo with psr4 folder in iban bic anonymizer

shakaran:
In composer update, fix the following error:

```
Generating optimized autoload files
Class MakinaCorpus\DbToolsBundle\Anonymization\Anonymizer\Common\IbanBicAnonymizer located 
in ./vendor/makinacorpus/db-tools-bundle/src/Anonymization/Anonymizer/Core/IbanBicAnonymizer.php 
does not comply with psr-4 autoloading standard. Skipping.
```

It seems a typo using "Common" folder which not exist, instead use the "Core" folder which have all the classes.


SimonMellerin (approved):
Thanks !

Issue #102, opened 2024-02-26
Make backup and restoration timeouts configurable


Lonnytunes:
The timeout of backup and restoration tasks should be configurable in the `db_tools.yaml` file.

PR #151, opened 2024-05-08
issue #102 - Added configurable timeouts for backup and restore

picturestone:


picturestone:
This MR makes the timeouts for the backup and restore processes configurable (related issue #102). 

I've run the `./dev.sh checks` and `./dev.sh test_all` commands as described in https://dbtoolsbundle.readthedocs.io/en/stable/contribute/guide.html and they run through, though some tests are incomplete or skipped. However, i've also run the same commands on the main branch and the same tests are skipped/incomplete on both main and my issue-102-configurable-timeouts branch. If the test executions were okay before, they are also okay now.

pounard:
@picturestone first thank you very musch ! It looks good to me, although I'm looking it with my phone I can't do a thourough review right now. I'll do in the next few days. @SimonMellerin I'll wait for your feedback as well!

src/Backupper/AbstractBackupper.php
```diff
@@ -138,9 +139,21 @@ public function isVerbose(): bool
         return $this->verbose;
     }
 
+    public function setTimeout(float $timeout): self
```
pounard, line 142: Why float if the original value is an integer ?

src/Backupper/AbstractBackupper.php
```diff
@@ -138,9 +139,21 @@ public function isVerbose(): bool
...
+    {
+        $this->timeout = $timeout;
+
+        return $this;
+    }
+
+    public function getTimeout(): float
```
pounard, line 149: Same as upper.

config/packages/db_tools.yaml
```diff
@@ -28,6 +28,12 @@ db_tools:
     # (Use relative date/time formats : https://www.php.net/manual/en/datetime.formats.relative.php)
     #backup_expiration_age: '6 months ago' # default '3 months ago'
 
+    # Timeout for backups.
+    # backup_timeout: 1200 # default 600
```
pounard, line 32: I would love to see this accepting date string such as "3 minutes" for example, this can be implemented by validating it using `\DateInterval::createFromDateString()` when not `\ctype_digit()` in the extension class.

src/DependencyInjection/DbToolsExtension.php
```diff
@@ -37,10 +37,12 @@ public function load(array $configs, ContainerBuilder $container): void
         $container->setParameter('db_tools.backupper.options', $config['backupper_options']);
         $container->setParameter('db_tools.backup_expiration_age', $config['backup_expiration_age']);
         $container->setParameter('db_tools.excluded_tables', $config['excluded_tables'] ?? []);
+        $container->setParameter('db_tools.backup_timeout', $config['backup_timeout']);
```
pounard, line 40: May be values should be validated here (not negative or not zero for example).

src/Command/BackupCommand.php
```diff
@@ -26,15 +26,18 @@ class BackupCommand extends Command
...
     public function __construct(
         string $defaultConnectionName,
         private BackupperFactory $backupperFactory,
         private Storage $storage,
+        float $timeout,
     ) {
         parent::__construct();
 
         $this->connectionName = $defaultConnectionName;
+        $this->timeout = $timeout;
```
SimonMellerin, line 40:
```suggestion=
    public function __construct(
        string $defaultConnectionName,
        private BackupperFactory $backupperFactory,
        private Storage $storage,
        private float $timeout,
    ) {
        parent::__construct();

        $this->connectionName = $defaultConnectionName;
```

SimonMellerin: We try to specify private properties in constructor to avoid boilerplate code as much as possible :)

src/Command/RestoreCommand.php
```diff
@@ -26,15 +26,18 @@ class RestoreCommand extends Command
...
         string $defaultConnectionName,
         private RestorerFactory $restorerFactory,
         private Storage $storage,
+        float $timeout,
     ) {
         parent::__construct();
 
         $this->connectionName = $defaultConnectionName;
+        $this->timeout = $timeout;
     }
```
SimonMellerin, line 41: same as above 

```suggestion

    public function __construct(
        string $defaultConnectionName,
        private RestorerFactory $restorerFactory,
        private Storage $storage,
        private float $timeout,
    ) {
        parent::__construct();

        $this->connectionName = $defaultConnectionName;
    }
```

SimonMellerin (commented):
Thanks for your work :pray: 

For this to be complete, a piece of documentation could be nice in `docs/content/configuration.md` :)

src/Backupper/AbstractBackupper.php
picturestone: I chose float because the symfony process we set the timeout for also uses a float

src/DependencyInjection/DbToolsExtension.php
picturestone: The symfony process also validates the timeout. I would explicitly allow 0 to let a process run indefinitely if the user so wishes.

```
    /**
     * Validates and returns the filtered timeout.
     *
     * @throws InvalidArgumentException if the given timeout is a negative number
     */
    private function validateTimeout(?float $timeout): ?float
    {
        $timeout = (float) $timeout;

        if (0.0 === $timeout) {
            $timeout = null;
        } elseif ($timeout < 0) {
            throw new InvalidArgumentException('The timeout value must be a valid positive integer or float number.');
        }

        return $timeout;
    }
```

pounard: OK, I have numerous comments about this piece of code:

- First, it should be a `int`, not a `float`.
- `float` comparison is dangerous, because of float imprecision, so you should not check against 0.0, but instead `if (null === $timeout) { /* ... */ } else if ($timeout <= 0) { throw ... }`

src/Backupper/AbstractBackupper.php
pounard: Oh, I see, I don't have a strong opinion then, but I'd still prefer to keep int and keep seconds as the lowest unit. It's backup and restore operations and they will probably never run under a second. What's your opinion about this @SimonMellerin ?

SimonMellerin: Ok I understand you choice but perhaps we can deviate from the behavior of Symfony Process Component here. 

The Symfony Process Component addresses general purpose  process than can be speeder than a second. But as @pounard said, our use case is a more specific and we will never need to be so precise with this timeout.

This being said, I found it easier to understand to have an integer for this configuration.

src/DependencyInjection/DbToolsExtension.php
picturestone: Hello again, this is not my code, this is code from the symfony repository: https://github.com/symfony/process/blob/820da2d703f2904ae04bbee634af705febfc1d68/Process.php#L1351
i'm not gonna take it upon me to change things there

i just copied this code into this comment to tell you that it is validated by the process itself, so i dont see why we should validate it to be a certain way as well, but if you want so i can add some validation to your repo for the timeout as well.

I however disagree with the timeout needing to be bigger than 0, because in pretty much any configuration you can set timeouts to 0 for things you want to run indefinitely (e.g. php's set_time_limit)

src/Command/BackupCommand.php
picturestone: I see, should i also change it for `$this->connectionName = $defaultConnectionName;`? since that is not my code, that was already there.

Or was this done on purpose because its important that the command has the parameter `defaultConnectionName`, but the property is named `connectionName `?

pounard: I think the `$defaultConnectionName` might be wrong in this context, I'm OK with naming the parameter `$connectionName`.

src/DependencyInjection/DbToolsExtension.php
pounard: Yes that's true, but I do prefer things to be explicit, i.e. `null` means no timeout, a value means there is one, it's pretty much the convention we took everywhere else in this project (there might be exceptions thought). @SimonMellerin what's your opinion ?

picturestone: so "null" as a string should lead to no timeout? because no value (so "null" for real) in the config currently leads to the default 600 which is the hardcoded timeout right now. I would leave it at "600 if no value provided" for backwards compatibility reasons

pounard: Validation and string to float conversion must be done in container compilation phase (or more generally when bootstrapping the configuration or services). We have plans in our roadmap to make this package runnable as a standalone app (without Symfony) so I wouldn't be against a proper reusable validation method somewhere.

Considering this, it would be great to add a `public static parseTimeout(null|int|float|string): ?int` method, which proceed to _user input_ validation I described upper (using `ctype_digit()` and `\DateInterval::createFromDateString()`). But this probably does not belong to this class. This method should also validate other constraints (more than 0 or null) even thought it would be duplicated work, because the validation also is required in the process builder as well (nothing prevent an API user from using it directly without using a validated user input).

This method could be written in a new `Validation` class within the `Helper` namespace.

This method can then be called when parsing configuration in the container extension, for validating the user input, and later in the future will be used in the standalone application bootstrap as well.

We'd very much like to keep an int here for the timeout, which will eventually be cast as float when passing it to Symfony process component when spawning it.

I'm OK with validating 0 as being considered "no timeout", but in order to be consistent with our code conventions, please consider setting `null` value to the class property when validating.

picturestone: i would suggest leaving 0 as "no timeout", otherwise handling the parameter could unnecessarily complicated together with the wish for a date interval from string to also work. Any programmer who has at least some experience with timeouts will assume that 0 means infinite, as it seems to be convention to me.

pounard: > i would suggest leaving 0 as "no timeout",

This goes against about usual conventions of this project, but if you feel it to be better, I'm OK with an non-nullable `int` property whose `0` value means _no timeout_. This is a minor detail and I don't wish it to block this PR !

pounard:
Hello @picturestone !

We have currently a few other PR that are awaiting for this one to get in, because we'd like to pass your fix in the next point release.

So, I took the liberty of forking your own branch and add a commit with all the suggestions we were expecting from you, so we may proceed with the next release in the next few days.

You can see it there: #161. I'm closing this PR. Since I kept your commits in my own branch, you will be credited in the main repository.

Thank you very much again for the work, we appreciate it very much!

PR #152, opened 2024-05-24
Standalone

pounard:
Standalone console tool, work in progress.

- [x] basic services bootstrap #153
- [x] dedicated configuration #153
- [x] standalone console application #153
- [x] configuration files automatic lookup #153
- [x] ~~anonymization dumping command (symfony side)~~ #156
- [x] ~~environment variables~~ will be done in #162
- [ ] tests, tests, tests
- [x] ~~integration into dev.sh script for running it~~ Not really useful, will see that later.
- [x] configuration files merging #153
- [x] relative path lookup relative to configuration files #153
- [x] PHAR file creation #156


src/Bridge/Standalone/Bootstrap.php
```diff
@@ -0,0 +1,440 @@
...
+        );
+    }
+
+    /**
+     * Gets the application root dir (path of the project's composer file).
+     */
+    private static function getProjectDir(LoggerInterface $logger): ?string
+    {
+        // 4 level of \dirname() gets us in this project parent folder.
+        $candidates = [\getcwd(), \dirname(__DIR__, 4)];
```
pounard, line 205: Note for myself: this doesn't work when run as a PHAR file.

src/Bridge/Standalone/Bootstrap.php
```diff
@@ -0,0 +1,440 @@
...
+            foreach ((array) $input->getOption('config') as $filename) {
+                $configFiles[] = $filename;
+            }
+        }
+
+        try {
+            $logger = new ConsoleLogger($output);
+            $context = self::bootstrap($config, $configFiles, $logger);
+            $defaultConnectionName = $context->databaseSessionRegistry->getDefaultConnectionName();
+        } catch (ConfigurationError $e) {
```
pounard, line 92: This is wrong because this prevents the default `list` command of composer, and disallow options usage for choosing the config file.

Note for myself: lazy-initialize the configuration only once the commands get executed.

src/DbToolsBundle.php
```diff
@@ -4,15 +4,27 @@
...
+ *   This class is deprecated and will be remove in next major.
+ * @see \MakinaCorpus\DbToolsBundle\Bridge\Symfony\DbToolsBundle
+ */
+class DbToolsBundle extends TheRealBundle
 {
     #[\Override]
     public function build(ContainerBuilder $container): void
     {
-        $container->addCompilerPass(new DbToolsPass());
+        \trigger_deprecation(
```
pounard, line 20: This reminds me that we have to test how the PHAR compiled tool will behave when used in PHP 8.1 (we must ensure that various polyfills are integrated into the PHAR file).

pounard:
Fixes #153, #154, #155.

composer.json
```diff
@@ -13,26 +13,38 @@
...
         "symfony/process": "^6.0|^7.0",
         "symfony/yaml": "^6.0|^7.0"
     },
     "require-dev": {
+        "doctrine/doctrine-bundle": "^2.10.0",
         "doctrine/orm": "^2.15|^3.0",
         "friendsofphp/php-cs-fixer": "^3.34",
         "phpstan/phpstan": "^1.10",
         "phpunit/phpunit": "^10.4",
+        "symfony/console": "^6.0|^7.0",
```
SimonMellerin, line 31: This is not only necessary on dev, no?

composer.json
```diff
@@ -13,26 +13,38 @@
...
         "symfony/framework-bundle": "^6.0|^7.0",
+        "symfony/password-hasher": "^6.0|^7.0",
         "symfony/validator": "^6.3|^7.0"
     },
+    "suggest": {
+        "composer/pcre": "For PHAR compilation",
+        "doctrine/doctrine-bundle": "For autoconfiguration in Symfony project context",
+        "seld/phar-utils": "For PHAR compilation",
+        "symfony/password-hasher": "In order to use the password hash anonymizer"
+    },
```
SimonMellerin, line 42:
```suggestion
    "suggest": {
        "doctrine/doctrine-bundle": "For autoconfiguration in Symfony project context",
        "symfony/password-hasher": "In order to use the password hash anonymizer"
    },
```

I think phar compilation will be, at the end, made by a CI job and will only concern a few people. Thus I wouldn't suggest installing these dependencies.

composer.json
```diff
@@ -13,26 +13,38 @@
...
+    "conflict": {
+        "composer/pcre": "<3.1|>=4.0",
+        "seld/phar-utils": "<1.2|>=2.0",
+        "symfony/password-hasher": "<6.0|>=8.0"
```
SimonMellerin, line 46: May be we should add some restriction with the doctrine/doctrine-bundle here ? Or on doctrine/orm ?

config/db_tools.standalone.sample.yaml
```diff
@@ -0,0 +1,63 @@
+# This configuration file is an example for standalone usage.
+#
+# The given parameters you will find in this file must not be set in a Symfony
+# application context, and will ignored if so.
```
SimonMellerin, line 4:
```suggestion
# application context, and will be ignored if so.
```

config/db_tools.standalone.sample.yaml
```diff
@@ -0,0 +1,63 @@
...
+# Working directory is the path to which all relative file references will
+# be relative to. If none set, the path will be this file directory instead.
+workdir: /var/www/my_project/
+
+# Database connections.
+# One line per connection, a single database URL, all options as query
+# parameters. Connections will be made using makincorpus/query-builder
+# which will raise exceptions when invalid options are found.
+# There is less configuration amplitude than using doctrine/dbal in
+# Symfony, yet it should be enough in most case.
```
SimonMellerin, line 19: Is there a documentation somewhere on how to write those database URL ?

config/db_tools.standalone.sample.yaml
```diff
@@ -0,0 +1,63 @@
...
+# In case any options or specific behaviour is missing, please file
+# an issue at https://github.com/makinacorpus/php-query-builder/issues
+connections:
+    connection_one: "pgsql://username:password@hostname:port?version=16.0&other_option=..."
+    connection_two: "mysql://username:password@hostname:port?version=8.1&other_option=..."
+
```
SimonMellerin, line 25: is it possible to just write 

`connections: "pgsql://username:password@hostname:port?version=16.0&other_option=..."`

if i have only one to declare ?

config/db_tools.standalone.sample.yaml
```diff
@@ -0,0 +1,63 @@
...
+# You can explicitely set which will be default connection in use when
+# none providen in the command line options. If you omit this configuration
+# vlaue, then the first one in the list will be used.
```
SimonMellerin, line 28:
```suggestion
# value, then the first one in the list will be used.
```

config/db_tools.standalone.sample.yaml
```diff
@@ -0,0 +1,63 @@
...
+default_connection: connection_one
+
+# Storage is autoconfigured when using Symfony, but when using this
+# application standalone, it's on your own. You must provide at least
+# a root directory for backups.
```
SimonMellerin, line 33:
```suggestion
# Using the DbToolsBundle standalone, you must provide at least
# a root directory for backups.
```

config/db_tools.standalone.sample.yaml
```diff
@@ -0,0 +1,63 @@
...
+storage:
+    # Path can be relative, it will be relative to the configuration file.
```
SimonMellerin, line 35: why not relative to the workdir declared up above ?

config/db_tools.standalone.sample.yaml
```diff
@@ -0,0 +1,63 @@
...
+    # If none provided, the default will be the following one.
+    root_dir: ./var/db_tools
+
+# From here, all options in the 'db_tools' section of Symfony configuration
+# can be added, including anonymization files configuration.
```
SimonMellerin, line 40: Maybe we can directly copy/paste the whole file here. 

I imagine that there could have user that don't now at all what is Symfony (ok may be not, but you know what i mean) and don't want to look for this file.

config/db_tools.standalone.sample.yaml
```diff
@@ -0,0 +1,63 @@
...
+    # From here you can proceed with manual file inclusion. Pathes can be
+    # either relative or absolute. Relative paths are relative to and resolved
+    # from this configuration file.
+    yaml:
+        connection_one: ./db_tools.anonymizer.connection_one.yaml
+        # ... other connections ...
+
+    # Extra configuration options, if you don't want to split the anonymization
+    # configuration into multiple files, you can directly write it here. If you
+    # use the entity dump capability, you can copy paste the result directly
```
SimonMellerin, line 52: I think it's to soon to talk about this here.

```suggestion
    # configuration into multiple files, you can directly write it here.
```

src/Anonymization/Anonymizer/Core/PasswordAnonymizer.php
```diff
@@ -22,6 +23,8 @@ class PasswordAnonymizer extends AbstractAnonymizer
     #[\Override]
     protected function validateOptions(): void
     {
+        MissingDependencyException::check('symfony/password-hasher', PasswordHasherFactory::class);
```
SimonMellerin, line 26: Doesn't PHP grumble right from `PasswordHasherFactory::class` if the component is missing ?

src/Bridge/Standalone/Bootstrap.php
```diff
@@ -0,0 +1,466 @@
...
+use MakinaCorpus\DbToolsBundle\Anonymization\Anonymizer\AnonymizerRegistry;
+use MakinaCorpus\DbToolsBundle\Backupper\BackupperFactory;
+use MakinaCorpus\DbToolsBundle\Command\Anonymization\AnonymizeCommand;
+use MakinaCorpus\DbToolsBundle\Command\Anonymization\AnonymizerListCommand;
+use MakinaCorpus\DbToolsBundle\Command\Anonymization\CleanCommand;
+use MakinaCorpus\DbToolsBundle\Command\Anonymization\ConfigDumpCommand;
+use MakinaCorpus\DbToolsBundle\Command\BackupCommand;
+use MakinaCorpus\DbToolsBundle\Command\CheckCommand;
+use MakinaCorpus\DbToolsBundle\Command\RestoreCommand;
+use MakinaCorpus\DbToolsBundle\Command\StatsCommand;
```
SimonMellerin, line 18:
```suggestion
use MakinaCorpus\DbToolsBundle\Command;
```

src/Bridge/Standalone/Bootstrap.php
```diff
@@ -0,0 +1,440 @@
...
+
+    /**
+     * Create anonymizer registry and register custom code and additional packs.
+     */
+    private static function createAnonymizeRegistry(array $config): AnonymizerRegistry
+    {
+        $projectDir = null;
+        $paths = [];
+
+        // @todo find a way to register packs when not in a composer project
```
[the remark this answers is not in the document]
pounard: I forgot to push, now it's like this:

```php
    /**
     * Create anonymizer registry and register custom code and additional packs.
     */
    private static function createAnonymizeRegistry(array $config): AnonymizerRegistry
    {
        // @todo find a way to register packs when not in a composer project

        return new AnonymizerRegistry($config['workdir'], $config['anonymizer_paths']);
    }
```

composer.json
pounard: Yes, that's a good point. I'm going to remove this ASAP.

pounard: Done.

config/db_tools.standalone.sample.yaml
pounard: It's documented in the configuration reference page, it's also possible to write `connections: URL` directly. I can add it here as well.

composer.json
pounard: Sorry I misread, I thought it was about the doctrine bundle...

composer.json
pounard: `makinacorpus/query-builder` already handles `doctrine/dbal` and that's the whole point of delegating connection handling to it. Although, we still have one single class where we use it directly: in the database session manager. I think it's OK to add restrictions, but let's be very open with it, such as `^3|^4` for DBAL and `^2|^3` for ORM.

pounard: Done.

config/db_tools.standalone.sample.yaml
pounard: I'm going to add this here, but better documentation should come into #162.

config/db_tools.standalone.sample.yaml
pounard: Applied.

config/db_tools.standalone.sample.yaml
pounard: Applied.

config/db_tools.standalone.sample.yaml
pounard: `workdir` is optional, per default it's the current directory you run the command from.

The lookup algorithm is the following:
 - if workdir is specified, it becomes the path on which all files will be relative,
 - if not, files will be relative to the configuration file itself.

This is the best I could come up with, and I think it's quite good because you can run the command from anywhere without having the behavior change.

Maybe what's disturbing you is the fact it's not documented at this place ?

pounard: I rewrote this paragraph.

config/db_tools.standalone.sample.yaml
```diff
@@ -0,0 +1,63 @@
...
+    # If none provided, the default will be the following one.
+    root_dir: ./var/db_tools
+
+# From here, all options in the 'db_tools' section of Symfony configuration
+# can be added, including anonymization files configuration.
+
+anonymization:
+    # From here you can proceed with manual file inclusion. Pathes can be
+    # either relative or absolute. Relative paths are relative to and resolved
+    # from this configuration file.
```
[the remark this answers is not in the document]
pounard: Same thing, per default workdir overrides the file directory, it's written on top of the file.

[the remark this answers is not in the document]
pounard: Rephrased as well.

config/db_tools.standalone.sample.yaml
pounard: Applied.

src/Anonymization/Anonymizer/Core/PasswordAnonymizer.php
pounard: Nope, class names are handled OK, it will crash only if you try to use it (new instance, static method, constant...), but the `Foo::class` expression is just a string, containing the whole class name as found in the `use` statement.

src/Bridge/Standalone/Bootstrap.php
```diff
@@ -0,0 +1,466 @@
...
+     * Create Symfony console application.
+     */
+    public static function createApplication(): Application
+    {
+        // @todo Test in PHAR context.
+        if (\class_exists(InstalledVersions::class)) {
+            $version = InstalledVersions::getVersion('makinacorpus/db-tools-bundle');
+        }
+        $version ??= 'cli';
+        \assert($version !== null);
```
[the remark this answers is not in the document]
pounard: Indeed, that's useless.

src/Bridge/Standalone/Bootstrap.php
```diff
@@ -0,0 +1,466 @@
...
+    {
+        static::createApplication()->run();
+    }
+
+    /**
+     * Create Symfony console application.
+     */
+    public static function createApplication(): Application
+    {
+        // @todo Test in PHAR context.
```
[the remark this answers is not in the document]
pounard: Nope actually it works great.

src/Bridge/Standalone/Bootstrap.php
```diff
@@ -0,0 +1,466 @@
...
+        if (\class_exists(InstalledVersions::class)) {
+            $version = InstalledVersions::getVersion('makinacorpus/db-tools-bundle');
+        }
+        $version ??= 'cli';
+        \assert($version !== null);
+
+        $application = new Application('Db Tools', $version);
```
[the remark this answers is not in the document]
pounard: In other places you simple wrote "DbTools", which one is correct? Here we're not a bundle but a standalone CLI app.

src/Bridge/Standalone/Bootstrap.php
```diff
@@ -0,0 +1,466 @@
...
+
+        // We need to parse a few arguments prior running the console
+        // application in order to setup commands. This is hackish but
+        // should work.
+        $config = $configFiles = [];
+        if ($input->hasOption('config')) {
+            foreach ((array) $input->getOption('config') as $filename) {
+                $configFiles[] = $filename;
+            }
+        }
```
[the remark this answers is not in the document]
pounard: Yes it is.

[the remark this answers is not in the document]
pounard: And it's not overriding anything, it simply use the given configuration file, and no other (there is no automatic resolution if you provide a file).

src/Bridge/Standalone/Bootstrap.php
```diff
@@ -0,0 +1,466 @@
...
+                fn (Context $context) => new StatsCommand(
+                    defaultConnectionName: $context->databaseSessionRegistry->getDefaultConnectionName(),
+                    statsProviderFactory: $context->statsProviderFactory,
+                ),
+                'Give some database statistics',
+                ['stats'],
+            ],
+        ];
+
+        $initializer = static fn (): Context => self::bootstrap($config, $configFiles, new ConsoleLogger($output));
```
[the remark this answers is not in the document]
pounard: That's remnant from the first iterations I guess, the idea was to let the early console bootstrap some latitude and pre-configure a few things. Also the PHP code is public and can be called by other framework integrations not using Symfony, case in which it allows those future integrations passing a few presets.

src/Bridge/Standalone/Bootstrap.php
```diff
@@ -0,0 +1,466 @@
...
+     *   Merged proper configuration.
+     */
+    private static function configParse(array $config, array $files, LoggerInterface $logger): array
+    {
+        $configFileNames = ['db_tools.yaml', 'db_tools.yml', 'db_tools.config.yaml', 'db_tools.config.yml'];
+        $projectRoot = self::getProjectDir($logger);
+        $workdir = $config['workdir'] ?? $projectRoot ?? \getcwd();
+
+        // When no configuration file given, attempt to find one.
+        // @todo Should stop at first when found.
```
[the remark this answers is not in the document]
pounard: Yes, it merges everything, thanks to the `symfony/config` component it should work transparently. I have no idea really if that's going to be useful one day... In this particular case (automatic lookup) it should stop at first, I'm going to fix that.

src/Bridge/Standalone/Bootstrap.php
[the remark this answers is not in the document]
pounard: I fixed it, when you don't provide a file and the automatic lookup triggers, it stops on the fist one found. Nevertheless I kept the loop continuing until the end and emiting debug output for people that are debugging their env.

src/Bridge/Standalone/Bootstrap.php
```diff
@@ -0,0 +1,466 @@
...
+
+        // Use symfony/config and our bundle configuration, which allows us
+        // to use it fully for validation and merge.
+        $configuration = new StandaloneConfiguration();
+        $processor = new Processor();
+
+        $config = $processor->processConfiguration($configuration, $configs);
+
+        // Set a base directory for file and backup lookup.
+        $config['workdir'] ?? $workdir;
```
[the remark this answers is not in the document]
pounard: What... stupid me.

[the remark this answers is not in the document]
pounard: It is actually working because this line has nothing to do here: it would override the relative path resolution if the user don't specify a workdir. I simply removed it.

src/Bridge/Symfony/DbToolsBundle.php
```diff
@@ -0,0 +1,31 @@
...
+declare (strict_types=1);
+
+namespace MakinaCorpus\DbToolsBundle\Bridge\Symfony;
+
+use MakinaCorpus\DbToolsBundle\Bridge\Symfony\DependencyInjection\Compiler\DbToolsPass;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\HttpKernel\Bundle\Bundle;
+use MakinaCorpus\DbToolsBundle\Bridge\Symfony\DependencyInjection\DbToolsExtension;
+
+class DbToolsBundle extends Bundle
```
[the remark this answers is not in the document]
pounard: I'm not even sure this would be necessary. The bundle works with zero conf (per default it maps everything on doctrine configuration) the only required thing is to add the bundle into your `config/bundles.php`.

src/Bridge/Standalone/Bootstrap.php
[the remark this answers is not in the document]
SimonMellerin: yep you're right, DbTools in one word maybe ? :)

src/Bridge/Standalone/Bootstrap.php
```diff
@@ -0,0 +1,466 @@
...
+        }
+
+        // 0 is not a good index for extension, this fails for false and 0.
+        if (!($pos = \strrpos($filename, '.'))) {
+            throw new ConfigurationException(\sprintf("%s: file extension cannot be guessed.", $filename));
+        }
+        $ext = \substr($filename, $pos + 1);
+
+        $config = match ($ext) {
+            'json' => \json_decode(\file_get_contents($filename), true, 512, \JSON_THROW_ON_ERROR),
```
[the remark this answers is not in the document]
SimonMellerin: :shrug:  @Lonnytunes  maybe ?

config/db_tools.standalone.sample.yaml
pounard: OK but this won't be a simple copy/paste, it'll require a few adjustments to better fit the use case of standalone CLI.

src/Bridge/Standalone/Bootstrap.php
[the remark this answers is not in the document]
pounard: And it's fixed!

src/Bridge/Standalone/Bootstrap.php
```diff
@@ -0,0 +1,466 @@
...
+        // Set a base directory for file and backup lookup.
+        $config['workdir'] ?? $workdir;
+
+        return $config;
+    }
+
+    /**
+     * Parse a single configuration file.
+     */
+    private static function configParseFile(string $filename): array
```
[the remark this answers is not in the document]
pounard: I basically prefixed all config-related methods using `configDoSomething()` instead of `doSomethingToConfig()`, that's an arbitrary choice, but I'm OK going the other way around. Simply tell me, if you want to change, I'll have to rename all methods, that's easy one.

src/Bridge/Symfony/DbToolsBundle.php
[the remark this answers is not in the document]
pounard: OK problem lies in the fact that flex automatically searches for `<your PSR-4>[Bundle]/<your PSR-4>Bundle` named class, but if we put it deeper in `src/Bridge/Symfony` it'll simply don't work anymore... Since we kept a bundle class with the original name for backward compatibility, it's not going to be a problem before 3.0, but I've opened an issue: https://github.com/makinacorpus/DbToolsBundle/issues/168

src/Database/StandaloneDatabaseSessionRegistry.php
```diff
@@ -0,0 +1,59 @@
...
+
+use MakinaCorpus\DbToolsBundle\Error\ConfigurationException;
+use MakinaCorpus\QueryBuilder\BridgeFactory;
+use MakinaCorpus\QueryBuilder\DatabaseSession;
+use MakinaCorpus\QueryBuilder\Dsn;
+
+/**
+ * makinacorpus/query-builder based implementation.
+ */
+class StandaloneDatabaseSessionRegistry implements DatabaseSessionRegistry
```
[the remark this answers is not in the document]
pounard: And the Doctrine implementation in `Bridge/Symfony` ? or in `Bridge/Doctrine` ?

src/Bridge/Standalone/Bootstrap.php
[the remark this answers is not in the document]
SimonMellerin: it also suits me like that, keep it

src/Bridge/Standalone/Bootstrap.php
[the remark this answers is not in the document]
SimonMellerin: I'm ok to keep it, but only if it's unit tested :)

src/Database/StandaloneDatabaseSessionRegistry.php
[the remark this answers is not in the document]
SimonMellerin: Bridge/Symfony seems good to me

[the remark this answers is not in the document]
pounard: OK, doctrine is not related to symfony but symfony does integrate it tightly, nevertheless I don't see any major use case of doctrine without symfony, so I'm OK with it in the symfony bridge.

composer.json
pounard: OK, what I wanted to do is:

* remove `symfony/console` from mandatory requirements in order to have the less hardwired dependencies possible,
* move it do dev dependencies for testing,
* console component is required for Symfony and the standalone build
* console component may conflict with an hypothetical future Laravel command, since it's own symfony/console version requirement may differ from ours.

@SimonMellerin that's why I want to keep it optional, so that potential other integrations can use, not use it, or use it in a different version.

src/Bridge/Standalone/Bootstrap.php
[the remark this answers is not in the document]
pounard: I'm just going to remove it, but I'll keep the match with a single case and default that raise an exception. It paves the way for a future configuration format.

composer.json
SimonMellerin: I don't know, as all current uses requires the Symfony/Console component, I find it odd to not require it.

May be we could address this hypothetical issue when it shows up ? 

Cause if I understand correctly, what you suggest imply that if one wants to use the DbToolsBundle on a random composer project, he should enter `composer require makinacorpus/db-tools-bundle symfony/console`, that's it ?

If so, it's a shame

pounard: True, let's keep it mandatory for now, but we will have to think thoroughly about this when we'll diversify integrations. We will eventually hit a hard barrier were some X or Y framework will cause dependency conflicts.

SimonMellerin (approved):

PR #169, opened 2024-05-31
Fix deprecation conversion float to int

maxhelias:
Capture d’écran 2024-05-31 à 17 08 52


SimonMellerin (changes requested):
Hi,

First of all, thank you for your contribution.

We'd rather use an explicit floor cast for this one.

Also, as you modify these lines, I think it's more readable with multi-line .

Thanks

SimonMellerin (approved):

PR #170, opened 2024-06-03
Fix default anonymizer_paths in config

AntoineAresu:
Fix example in config file, with correct file path


SimonMellerin (approved):
Hi,

Thank you for the fix

PR #177, opened 2024-07-06
fix: Extension deprecation

Chris53897:
Will fix this deprecation-message

`The "Symfony\Component\HttpKernel\DependencyInjection\Extension" class is considered internal since Symfony 7.1, to be deprecated in 8.1; use Symfony\Component\DependencyInjection\Extension\Extension instead. It may change without further notice. You should not use it from "MakinaCorpus\DbToolsBundle\DependencyInjection\DbToolsExtension".`


SimonMellerin (approved):
Thanks

PR #179, opened 2024-09-20
Update coding-standards.yml

Chris53897:


pounard:
I would like very much to have some kind of discussions and politeness chattering whenever this kind of PR happens. But, I made the same change in another package and it's fine, I'm OK to merge it. ping @SimonMellerin

SimonMellerin:
I totally agree with @pounard , besides, FYI, we will soon update our contribution rules to make it mandatory to associate each PR with an issue. 

Like this, we will prevent wide PR like this one and it will make our maintenance work much easier.

SimonMellerin (approved):

PR #181, opened 2024-09-20
chore: Add Tests for PHP 8.4

Chris53897:


pounard:
Thanks @Chris53897 - I'm closing this intentionally and opening an issue instead, since we have a custom dev script, which needs to be updated as well, we are waiting for official php 8.4 and xdebug 3.4 releases for enabling testing at once in CI and our dev script.

pounard:
Opened https://github.com/makinacorpus/DbToolsBundle/issues/186

PR #198, opened 2024-11-22
feature: fakerphp/faker integration

pounard:
Two new anonymizers:
 - First one, an address anonymizer, but using faker. 
 - Second one, an "any faker method" anonymizer, which awaits a `method` option, that will then dynamically call this method name on the faker generator for building the sample table.

I also took the liberty of refactoring the sample size handling, in order for pretty much all anonymizers to be able to use it.

The `fakerphp/faker` dependency is optional, in `suggests`, in order to make it work gracefully, I added a warning when a dependency is missing in `anonymization:list` command.

### DIFF
--- a/src/Anonymization/Anonymizer/AbstractAnonymizer.php
+++ b/src/Anonymization/Anonymizer/AbstractAnonymizer.php
@@ -110,7 +110,41 @@ abstract class AbstractAnonymizer
      *
      * @throws \Exception if any option is invalid.
      */
-    protected function validateOptions(): void {}
+    protected function validateOptions(): void
+    {
+        if ($this->hasSampleSizeOption()) {
+            if ($this->options->has('sample_size')) {
+                $value = $this->options->getInt('sample_size');
+                if ($value <= 0) {
+                    throw new \InvalidArgumentException("'sample_size' option must be a positive integer.");
+                }
+            }
+        }
+    }
+
+    /**
+     * Does this anonymizer has a "sample size" option.
+     */
+    protected function hasSampleSizeOption(): bool
+    {
+        return false;
+    }
+
+    /**
+     * Default sample size, goes along the "sample size" option set to true.
+     */
+    protected function getDefaultSampleSize(): int
+    {
+        return 500;
+    }
+
+    /**
+     * Default sample size, goes along the "sample size" option set to true.
+     */
+    protected function getSampleSize(): int
+    {
+        return $this->options->getInt('sample_size', $this->getDefaultSampleSize());
+    }
 
     /**
      * Initialize your anonymizer.
--- a/composer.json
+++ b/composer.json
@@ -35,6 +35,8 @@
         "symfony/validator": "^6.3|^7.0"
     },
     "suggest": {
+        "db-tools-bundle/pack-faker": "FakerPHP/Faker bridge anonyzer pack",
+        "db-tools-bundle/pack-fr-fr": "French anonymizers pack",
         "symfony/password-hasher": "In order to use the password hash anonymizer"
     },
     "conflict": {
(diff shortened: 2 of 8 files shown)


src/Anonymization/Anonymizer/AbstractAnonymizer.php
```diff
@@ -110,7 +110,41 @@ protected function getSalt(): string
...
+        return 500;
+    }
+
+    /**
+     * Default sample size, goes along the "sample size" option set to true.
+     */
+    protected function getSampleSize(): int
+    {
+        return $this->options->getInt('sample_size', $this->getDefaultSampleSize());
+    }
```
SimonMellerin, line 147: Why not put that in `AbstractEnumAnonymizer` ?

src/Anonymization/Anonymizer/Faker/FakerAddressAnonymizer.php
```diff
@@ -0,0 +1,97 @@
...
+    Map columns for each part with options ('country', 'locality', 'region', 'postal_code', 'street_address', 'secondary_address').
+    You must set the 'locale' parameter (ex. 'fr_FR', 'en_US', 'de_AT', ...).
+    You can also specify the sample table size using the 'sample_size' option
+    (default is 500). The more samples you have, the less duplicates you will
+    end up with.
+    TXT,
+    requires: [Faker\Factory::class],
+    dependencies: ['fakerphp/faker'],
+)]
+class FakerAddressAnonymizer extends AbstractMultipleColumnAnonymizer
```
SimonMellerin, line 25: Not sure we need another specific address anonymizer, is there a way to create a generic mutlicolumn faker anonymizer instead ?

SimonMellerin (changes requested):
That's a super good feature!

But not sure I want to see Faker arrives in core.

At the beginning, I wanted the DbToolsBundle to be modular precisely to avoid having a big lib as Faker.

It's something that Faker's creator himself regretted : it's something that Faker's creator himself regretted

So to keeping this original goal in mind, I'd prefer to have a `FakerPack` instead.

src/Anonymization/Anonymizer/AbstractAnonymizer.php
pounard: So that the multicolumn anonymizer can have it!

src/Anonymization/Anonymizer/Faker/FakerAddressAnonymizer.php
pounard: Actually the one using Faker is much better, because it supports tens of languages! I think we should probably consider recommend the one using Faker than our own.

pounard:
> At the beginning, I wanted the DbToolsBundle to be modular precisely to avoid having a big lib as Faker.
> 
> It's something that Faker's creator himself regretted : it's something that Faker's creator himself regretted
> 
> So to keeping this original goal in mind, I'd prefer to have a `FakerPack` instead.

I thought as well at the beginning, but the fact that having addresses in as much languages as Faker supports is really nice to have.

Further, if later we package a docker image or a phar archive, it would probably be completely useless without many anonymizers included!

The more I think of it, the more I think we will never manage supporting as many languages as Faker, and even more, I'd say that we don't even speak more than 3 languages ourselves.

pounard:
OK, following the discussion, I created https://github.com/DbToolsBundle/pack-faker which is not released yet, nor propagated on packagist.

It's awaiting review as well.

It nevertheless uses some internal changes I introduced in this very specific PR, so I'll need a review for it.

SimonMellerin (approved):

src/Anonymization/Anonymizer/Faker/FakerAddressAnonymizer.php
pounard: Close, faker integration is now into its own package.

Issue #194, opened 2024-11-20
Document configuration changes from #191


pounard:
Cf. title

PR #201, opened 2024-11-27
issue #194 - some more documentation

pounard:


docs/content/configuration/reference.md
```diff
@@ -29,22 +29,135 @@ relative to the configuration file directory the path is defined
...
+}
+</style>
+
+## `anonymization`
+
+Write the anonymization configuration directly in the main configuration file.
+
+Keys under the first dimension are connections names, then follows the structure
+expected in anonymization configuration files.
+
```
SimonMellerin, line 55: Maybe add something about that if there's only one connection and no name is given, the connection is called `default`

SimonMellerin (approved):

docs/content/configuration/reference.md
pounard: I added it in the `connections` configuration, but it may be repeated here.

Or maybe we should add an *Essentials* section on top, which explains how are handled all file paths, connections names, etc... ?

SimonMellerin: I don't know, I think that what make me sad here is that we have to precise `default` even if we only had one connection ^^

But there is no easy solution to this :grimacing:

docs/content/configuration/reference.md
```diff
@@ -289,8 +434,18 @@ connections:
...
 
+:::warning
+If you configure this parameter with a single URL string with no connection names,
+the connection name will be `default`.
+:::
+
 </div>
 
+:::tip
+All parameters for each connection are the exact same as the top-level parameters
```
Lonnytunes, line 445:
```suggestion
All parameters for each connection are exactly the same as the top-level parameters
```

docs/content/configuration/reference.md
```diff
@@ -289,8 +434,18 @@ connections:
 connections: "pgsql://username:password@hostname:port?version=16.0&other_option=..."
 ```
 
+:::warning
+If you configure this parameter with a single URL string with no connection names,
```
Lonnytunes, line 438:
```suggestion
If you configure this parameter with a single URL string with no connection name,
```

docs/content/configuration/reference.md
pounard: Hum, I just want to merge this. We are going to continue improving the documentation in the next few days or weeks.

Lonnytunes (approved):
Thanks for having considered my little suggestions :slightly_smiling_face:

docs/content/configuration/reference.md
pounard: I was adding a tip specifying this, but I realized I would have to add it everywhere we use connection names in configuration.

That's not that much sections, but it's a lot of copy/pasted pieces...

pounard: @SimonMellerin That's only 2, so I did it, is that OK for you ?

SimonMellerin: thanks !

Issue #178, opened 2024-09-16
Laravel Support


jihadadulrazaqjasim:
Is the package can be used in laravel ?

pounard:
Hello, this is a good question.

As of now, the core API is independent from any framework, and can be used as a standalone app (in `main` branch, which will be the next major 2.0 release). This means that you can manually write your database table and column configuration then run it, no matter the framework you use:

- You may start by reading the manual YAML configuration: https://dbtoolsbundle.readthedocs.io/en/stable/configuration.html#anonymization
- You can then attempt the `main` branch install, which includes a standalone CLI script: https://github.com/makinacorpus/DbToolsBundle/blob/main/bin/db-tools and run it manually.

Please note it might cause a few dependency problems with your own `composer.json` file, so you might want to try by installing it in its own project folder.

Laravel integration is something we discussed, and planned in our backlog, but with no due date. We would greatly appreciate some help on this by the way!

Lonnytunes:
Hi @jihadadulrazaqjasim,

You may have already received some notifications, but just to inform you that we are currently working on the Laravel support (!206).

Lonnytunes:
- [x] Base support with anonymization defined in configuration file(s) (#206)
- [ ] Anonymization definable in model classes
- [ ] Documentation
- [ ] Tests

PR #206, opened 2024-11-27
Issue #178 - Laravel support

Lonnytunes:
Base support with anonymization defined in configuration file(s).

### DIFF
--- a/src/Anonymization/Config/Loader/PhpFileLoader.php
+++ b/src/Anonymization/Config/Loader/PhpFileLoader.php
@@ -1,0 +1,40 @@
+<?php
+
+declare(strict_types=1);
+
+namespace MakinaCorpus\DbToolsBundle\Anonymization\Config\Loader;
+
+use MakinaCorpus\DbToolsBundle\Error\ConfigurationException;
+
+class PhpFileLoader extends ArrayLoader
+{
+    public function __construct(
+        private string $file,
+        string $connectionName = 'default',
+    ) {
+        parent::__construct([], $connectionName);
+    }
+
+    #[\Override]
+    protected function getData(): array
+    {
+        try {
+            $data = require $this->file;
+        } catch (\Throwable $e) {
+            throw new ConfigurationException(\sprintf(
+                "An error occurred when loading \"%s\" configuration file: %s",
+                $this->file,
+                $e->getMessage()
+            ), 0, $e);
+        }
+
+        if (!\is_array($data)) {
+            throw new ConfigurationException(\sprintf(
+                "File \"%s\" is not a valid PHP anonymization configuration file (must return an array).",
+                $this->file
+            ));
+        }
+
+        return $data;
+    }
+}
--- a/src/Bridge/Laravel/DbToolsServiceProvider.php
+++ b/src/Bridge/Laravel/DbToolsServiceProvider.php
@@ -1,0 +1,267 @@
+<?php
+
+declare(strict_types=1);
+
+namespace MakinaCorpus\DbToolsBundle\Bridge\Laravel;
+
+use Illuminate\Contracts\Config\Repository;
+use Illuminate\Contracts\Foundation\Application;
+use Illuminate\Foundation\Console\AboutCommand;
+use Illuminate\Support\ServiceProvider;
+use MakinaCorpus\DbToolsBundle\Anonymization\AnonymizatorFactory;
+use MakinaCorpus\DbToolsBundle\Anonymization\Anonymizer\AnonymizerRegistry;
+use MakinaCorpus\DbToolsBundle\Anonymization\Config\Loader\ArrayLoader;
+use MakinaCorpus\DbToolsBundle\Anonymization\Config\Loader\PhpFileLoader;
+use MakinaCorpus\DbToolsBundle\Anonymization\Config\Loader\YamlLoader;
+use MakinaCorpus\DbToolsBundle\Backupper\BackupperFactory;
+use MakinaCorpus\DbToolsBundle\Bridge\Symfony\DependencyInjection\DbToolsConfiguration;
+use MakinaCorpus\DbToolsBundle\Command\Anonymization\AnonymizeCommand;
+use MakinaCorpus\DbToolsBundle\Command\Anonymization\AnonymizerListCommand;
+use MakinaCorpus\DbToolsBundle\Command\Anonymization\CleanCommand;
+use MakinaCorpus\DbToolsBundle\Command\Anonymization\ConfigDumpCommand;
+use MakinaCorpus\DbToolsBundle\Command\BackupCommand;
+use MakinaCorpus\DbToolsBundle\Command\CheckCommand;
+use MakinaCorpus\DbToolsBundle\Command\RestoreCommand;
+use MakinaCorpus\DbToolsBundle\Command\StatsCommand;
+use MakinaCorpus\DbToolsBundle\Configuration\Configuration;
+use MakinaCorpus\DbToolsBundle\Configuration\ConfigurationRegistry;
+use MakinaCorpus\DbToolsBundle\Database\DatabaseSessionRegistry;
+use MakinaCorpus\DbToolsBundle\Error\ConfigurationException;
+use MakinaCorpus\DbToolsBundle\Restorer\RestorerFactory;
+use MakinaCorpus\DbToolsBundle\Stats\StatsProviderFactory;
+use MakinaCorpus\DbToolsBundle\Storage\DefaultFilenameStrategy;
+use MakinaCorpus\DbToolsBundle\Storage\FilenameStrategyInterface;
+use MakinaCorpus\DbToolsBundle\Storage\Storage;
+use Psr\Log\LoggerInterface;
+use Symfony\Component\Config\Definition\Processor;
+
+class DbToolsServiceProvider extends ServiceProvider
+{
+    #[\Override]
+    public function register(): void
+    {
+        $this->app->extend('config', function (Repository $config, Application $app) {
+            $dbToolsConfig = $config->get('db-tools', []);
+            $dbToolsConfig = ['db_tools' => $dbToolsConfig];
+
+            $processor = new Processor();
+            $configuration = new DbToolsConfiguration(false, true);
+
+            $dbToolsConfig = $processor->processConfiguration($configuration, $dbToolsConfig);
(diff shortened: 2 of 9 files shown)


src/Bridge/Laravel/DbToolsServiceProvider.php
```diff
@@ -0,0 +1,270 @@
...
+        /*$this->app
+            ->when([
+                AnonymizatorFactory::class,
+                BackupperFactory::class,
+                RestorerFactory::class,
+                StatsProviderFactory::class
+            ])
+            ->needs(LoggerInterface::class)
+            ->give(fn (Application $app) => $app->make('log'))
+        ;*/
```
SimonMellerin, line 224: why is this commented ?

src/Bridge/Laravel/DbToolsServiceProvider.php
```diff
@@ -0,0 +1,270 @@
...
+            ->giveConfig('database.default')
+        ;
+    }
+
+    /**
+     * Bootstrap any package services.
+     */
+    public function boot(): void
+    {
+        AboutCommand::add('DbToolsBundle', fn () => ['Version' => '2.0.0']);
```
SimonMellerin, line 251: may be you should add laravel or something as dev dependencies

cf the static analysis failure:

image

src/Anonymization/Config/Loader/PhpFileLoader.php
```diff
@@ -0,0 +1,32 @@
...
+        private string $file,
+        string $connectionName = 'default',
+    ) {
+        parent::__construct([], $connectionName);
+    }
+
+    #[\Override]
+    protected function getData(): array
+    {
+        $data = require $this->file;
```
pounard, line 21: No error handling here?

src/Bridge/Laravel/DbToolsServiceProvider.php
```diff
@@ -0,0 +1,270 @@
...
+        $this->app->extend('config', function (Repository $config, Application $app) {
+            $dbToolsConfig = $config->get('db-tools', []);
+            $dbToolsConfig = ['db_tools' => $dbToolsConfig];
+
+            $processor = new Processor();
+            $configuration = new DbToolsConfiguration(false, true);
+
+            $dbToolsConfig = $processor->processConfiguration($configuration, $dbToolsConfig);
+            $dbToolsConfig = DbToolsConfiguration::fixLegacyOptions($dbToolsConfig);
+            $dbToolsConfig = DbToolsConfiguration::appendPostConfig($dbToolsConfig);
```
pounard, line 47: I was thinking about this, maybe the `fixLegacyOptions()` method should be implicitly called within the `appendPostConfig()` method?

src/Bridge/Laravel/DbToolsServiceProvider.php
```diff
@@ -0,0 +1,270 @@
...
+        $this->app->singleton(DatabaseSessionRegistry::class, function (Application $app) {
+            return new LaravelDatabaseSessionRegistry($app->make('db'));
+        });
+
+        $this->app->singleton(ConfigurationRegistry::class, function (Application $app) {
+            /** @var Repository $config */
+            $config = $app->make('config');
+
+            $defaultConfig = new Configuration(
+                backupBinary: $config->get('db-tools.backup_binary', '/usr/bin/pg_dump'),
```
pounard, line 63: You should not set any default in configuration, the PHP classes will do their job on their own.

pounard: This is true for all variables, except for the `storage_directory` one.

src/Bridge/Laravel/DbToolsServiceProvider.php
```diff
@@ -0,0 +1,270 @@
...
+            );
+        });
+
+        $this->app->singleton(Storage::class, function (Application $app) {
+            /** @var ConfigurationRegistry $configRegistry */
+            $configRegistry = $app->make(ConfigurationRegistry::class);
+
+            $connections = $configRegistry->getConnectionConfigAll();
+            if (empty($connections)) {
+                $connections[$configRegistry->getDefaultConnection()] = $configRegistry->getDefaultConfig();
```
pounard, line 105: You don't need this if statement, the ConfigurationRegistry object does it for you when requested.

src/Bridge/Laravel/Resources/config/db-tools.php
```diff
@@ -0,0 +1,218 @@
...
+    | vendor. If you have more than one connection and if they use different
+    | database vendors or versions, please configure those for each connection
+    | instead.
+    |
+    | Default values depends upon vendor and are documented at
+    | https://dbtoolsbundle.readthedocs.io/en/stable/configuration.html
+    |
+    */
+
+    'backup_binary' => env('DBTOOLS_BACKUP_BINARY', '/usr/bin/pg_dump'),
```
pounard, line 103: Are you setting defaults here? You should not, the default bin and options to use are directly injected by the adapter when empty: if you set values in the configuration, it will override it and possibly create bugs later.

src/Bridge/Laravel/DbToolsServiceProvider.php
Lonnytunes: I wrote it before realizing it was probably not necessary. I did not remove it for now. I would like to be sure there is no interest in keeping this code.

src/Bridge/Laravel/DbToolsServiceProvider.php
Lonnytunes: Exactly what I wanted to avoid :sweat_smile: But it apparently won't be inevitable :slightly_smiling_face:

src/Bridge/Laravel/DbToolsServiceProvider.php
Lonnytunes: Ok! I will remove this `if` :ok_hand:

src/Bridge/Laravel/Resources/config/db-tools.php
Lonnytunes: I asked myself about that point. I actually defined default values in two places while thinking that I will have to keep them in only one. But considering your other related comment, I probably have to remove all default values I precised and rely on the `Configuration` class.

src/Bridge/Laravel/DbToolsServiceProvider.php
Lonnytunes: Wait, `ConfigurationRegistry` class does not handle this case. Method `getConnectionConfigAll()` may not be complete.

src/Anonymization/Config/Loader/PhpFileLoader.php
Lonnytunes: I will fix that!

src/Bridge/Laravel/DbToolsServiceProvider.php
Lonnytunes: Apparently `storage_directory` is not the only exception. For example, `Configuration` class has no default value for the backup binary, nor the backup options. Should I complete this class?

src/Bridge/Laravel/DbToolsServiceProvider.php
Lonnytunes: That seems relevant to me.

src/Bridge/Laravel/DbToolsServiceProvider.php
pounard: You dont need defaults for binaires, each backupper and restorer class provides its own.

src/Bridge/Laravel/DbToolsServiceProvider.php
pounard: I'm not sure, but I think that getConnectionConfigAll is not required for most commands

pounard: Now I'm sure, the getConnectionConfigAll() method only serves the purpose of knowing connections in the standalone bootstrap code, but in laravel, it's handled in the specific database session component instead.

src/Bridge/Laravel/DbToolsServiceProvider.php
Lonnytunes: Ok! I understand better now.

src/Bridge/Laravel/DbToolsServiceProvider.php
Lonnytunes: I looked at the `Storage` class more precisely and I realized that I actually can reduce this part of code :+1:

src/Bridge/Laravel/DbToolsServiceProvider.php
SimonMellerin: yep, that seems relevant to me too

src/Bridge/Laravel/DbToolsServiceProvider.php
SimonMellerin: Not sure to understand how it could work without it but I trust you ^^

SimonMellerin (approved):

src/Bridge/Laravel/DbToolsServiceProvider.php
Lonnytunes: @SimonMellerin https://laravel.com/docs/11.x/container#zero-configuration-resolution

Lonnytunes: @SimonMellerin As I let you know through the link above, Laravel will be able to instantiate classes in question in the commented code, since we tell it how to resolve their dependencies.

But if I did not misunderstand, Laravel will create a new instance each time an instance of those classes will be necessary.

I'm wondering if it is better that I restore the commented bindings to ensure that each instance will be a singleton.

What do you think?

SimonMellerin: Well I would say that those services was designed to be singleton (as it was first designed to be used in a symfony container). Thus, to avoid any side effect I think you are right. Go for singleton !

src/Bridge/Laravel/DbToolsServiceProvider.php
Lonnytunes: @SimonMellerin I renamed `appendPostConfig()` method to `finalizeConfiguration()`, since it is now intended to apply all types of modification we could have to do after the configuration file processing.

Issue #216, opened 2025-03-21
Symfony - `storage_filename_strategy` configuration not evaluated


MaSpeng:
The Symfony integration as of version 2.0 of this bundle does not evaluate the `storage_filename_strategy` and `connections.<connection_name>.storage.filename_strategy` configuration.

The service `db_tools.storage` will be configured with strategies resolved from the deprecated configuration path `db_tools.storage.filename_strategy` at https://github.com/makinacorpus/DbToolsBundle/blob/d0cb6322305a9a050917b70a2392fe31f4d83142/src/Bridge/Symfony/DependencyInjection/DbToolsExtension.php#L88-L109

Due to the initialization of the configuration object at https://github.com/makinacorpus/DbToolsBundle/blob/d0cb6322305a9a050917b70a2392fe31f4d83142/src/Bridge/Symfony/DependencyInjection/DbToolsExtension.php#L115 its not possible anymore to use this configuration path.

The configuration at `db_tools.storage_filename_strategy` and `db_tools.connections.<connection_name>.storage_filename_strategy` does not seem to be taken into account at all.

---

<details>
<summary>For anyone facing the same issue, as temporary a workaround, it is possible to update the definition of the service `db_tools.storage` through a custom compiler pass.</summary>

```php
<?php

declare(strict_types=1);

namespace App\DependencyInjection\Compiler;

use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
use Symfony\Component\DependencyInjection\ContainerBuilder;
use Symfony\Component\DependencyInjection\Exception\LogicException;
use Symfony\Component\DependencyInjection\Reference;

class DbToolsCompilerPass implements CompilerPassInterface
{
    public function process(ContainerBuilder $container): void
    {
        /**
         * @var array<int, array{
         *      storage_filename_strategy?: 'default'|'datetime'|string,
         *      connections: array<string, array{
         *          storage_filename_strategy?: 'default'|'datetime'|string,
         *      }
         *  } $dbToolsConfigurations
         */
        $dbToolsConfigurations = $container->getExtensionConfig('db_tools');
        if (count($dbToolsConfigurations) > 1) {
            throw new LogicException('More than one db tool configuration defined.');
        }

        $dbToolsConfiguration = $dbToolsConfigurations[0];
        if ([] === $dbToolsConfiguration) {
            return;
        }

        $storageFilenameStrategies = [];

[...]

PR #218, opened 2025-03-28
#216 fix symfony custom filename strategy

SimonMellerin:


Lonnytunes (approved):

src/Bridge/Symfony/DependencyInjection/DbToolsExtension.php
```diff
@@ -88,7 +88,7 @@ public function load(array $configs, ContainerBuilder $container): void
         $strategyServices = [];
         foreach ($config['connections'] as $connectionName => $connectionConfig) {
             // Default is handled directly by the storage service.
-            $strategyId = $connectionConfig['storage_filename_strategy'] ?? null;
+            $strategyId = $connectionConfig['storage_filename_strategy'] ?? $config['storage_filename_strategy'] ?? null;
```
pounard, line 91: This is not good, we have a configuration object for each user defined configuration, which falls back on a default configuration object at runtime when fetching its configuration. You should have checked that the default configuration is rightly populated.

PR #220, opened 2025-04-09
Anonymization - Mysql 8.0 - fix addAnonymizerIdColumnMySql

SimonMellerin:
'CREATE FUNCTION IF NOT EXISTS' is not strictly compatible with MYSQL 8.0 (beginning with MySQL 8.0.29)

### DIFF
--- a/src/Anonymization/Anonymizator.php
+++ b/src/Anonymization/Anonymizator.php
@@ -687,7 +687,7 @@ class Anonymizator implements LoggerAwareInterface
             } else {
                 $this->databaseSession->executeStatement(
                     <<<SQL
-                    CREATE FUNCTION IF NOT EXISTS ?::id() RETURNS BIGINT
+                    CREATE FUNCTION ?::id() RETURNS BIGINT
                     DETERMINISTIC
                     BEGIN
                         SELECT `value` + 1 INTO @value FROM ?::table LIMIT 1;


pounard:
This may crash if a previous run was stopped before cleaning, I'm not sure if this would really happen, but in theory, it could, and in this scenario, our users would be left with a very difficult problem to solve on their own.

src/Anonymization/Anonymizator.php
```diff
@@ -687,7 +687,7 @@ protected function addAnonymizerIdColumnMySql(string $table): void
             } else {
                 $this->databaseSession->executeStatement(
                     <<<SQL
-                    CREATE FUNCTION IF NOT EXISTS ?::id() RETURNS BIGINT
+                    CREATE FUNCTION ?::id() RETURNS BIGINT
```
Lonnytunes, line 690: I think we should drop the potential existing function before.

Lonnytunes (changes requested):
https://github.com/makinacorpus/DbToolsBundle/pull/220/files#r2034738628

src/Anonymization/Anonymizator.php
SimonMellerin: @Lonnytunes @pounard  there is this just before (L645)

```php
        $this->databaseSession->executeStatement(
            <<<SQL
            DROP FUNCTION IF EXISTS ?::id
            SQL,
            [$functionName],
        );
```

So it seems ok to me :)

Lonnytunes: So that's perfect :ok_hand: :slightly_smiling_face:

Lonnytunes (approved):

PR #221, opened 2025-07-18
Give anonymizer metadata responsibility to the AnonymizerRegistry class

pounard:
Preparation work for later, which will allow dynamic anonymizers to be created differently than from a class name. This allows dynamic anonymizers creation from configuration.

This will allow packs to be written without PHP code.

In order to allow this:
1. we move all anonymizer metadata code into the registry itself, it's not the anonymizer responsability anymore (since some will be dynamic, we cannot know them beforehand in order to create them then ask for metadata, this is a chicken and egg problem),
2. we move the responsibility to create instances to the registry as well, which then also acts as a factory.

This is a first step, tests won't pass right away since I didn't fix them all, it's a PoC for discussion.


src/Anonymization/Anonymizer/AnonymizerRegistry.php
```diff
@@ -45,42 +51,66 @@ public function addPath(array $paths): void
...
+        return new $className($config->table, $config->targetName, $databaseSession, $config);
+    }
+
+    /**
+     * Get anonymizer metadata.
+     */
+    public function getAnonymizerMetadata(string $name): AsAnonymizer
     {
-        $this->initialize();
+        return $this->metadata[$name] ?? throw new \InvalidArgumentException(\sprintf("Can't find Anonymizer with name : %s, check your configuration.", $name));
```
SimonMellerin, line 82:
```suggestion
        return $this->getAllAnonymizerMetadata()[$name] ?? throw new \InvalidArgumentException(\sprintf("Can't find Anonymizer with name : %s, check your configuration.", $name));
```

src/Anonymization/Anonymizer/AnonymizerRegistry.php
```diff
@@ -45,42 +51,66 @@ public function addPath(array $paths): void
...
-     *
-     * @return string
-     *   Anonymizer class name.
+     * Create anonymizer instance.
      */
-    public function get(string $name): string
+    public function createAnonymizer(
+        string $name,
+        AnonymizerConfig $config,
+        Options $options,
```
SimonMellerin, line 69: $options argument seems not to be used ? Why not remove it ?

SimonMellerin (commented):
except my 2 comments, seems good to me.

Just one question, are you sure that if we use the same anonymization twice (on 2 different tables and/or columns, it is instantiated only once ? (ie one sample table will be created and used)

src/Anonymization/Anonymizer/AnonymizerRegistry.php
```diff
@@ -26,7 +29,10 @@ class AnonymizerRegistry
         Core\StringAnonymizer::class,
     ];
 
-    private ?array $anonymizers = null;
+    /** @var array<string,string> */
+    private ?array $classes = null;
+    /** @var array<string,AsAnonymizer> */
+    private ?array $metadata = null;
```
Lonnytunes, line 35:
```suggestion
    /** @var array<string, string> */
    private ?array $classes = null;
    /** @var array<string, AsAnonymizer> */
    private ?array $metadata = null;
```

src/Anonymization/Anonymizer/AnonymizerRegistry.php
pounard: It is used actually, it's the last parameter of the anonymizer constructor, I might have accidentally forgotten it!

src/Anonymization/Anonymizer/AnonymizerRegistry.php
SimonMellerin: may be we should update phpcs-fixer rules!

src/Anonymization/Anonymizer/AnonymizerRegistry.php
pounard: I forgot the `$this->initialize();` call...

src/Anonymization/Anonymizer/AnonymizerRegistry.php
pounard: OK PHPstan documents it with whitespace, let's add those whitespaces.

src/Anonymization/Anonymizer/AnonymizerRegistry.php
pounard: Fixed!

SimonMellerin (approved):
ok for me, if we put aside the fact that the tests don't pass :sweat_smile:

pounard:
> ok for me, if we put aside the fact that the tests don't pass 😅

Yolo, the only way of living.

PR #224, opened 2025-07-22
feature: string pattern anonymizer, build complex strings by fetching values from other anonymizers

pounard:
See documentation for more information.

TL;DR:

```yaml
anonymization:
    default:
        customer:
            biography:
                anonymizer: pattern
                options: {value: '{firstname} {lastname} lives in {address:locality} in {address:country}, he has [2,7] cats.'}
```

## String range

By adding `[MIN,MAX]` anywhere in your text, it will be replaced by a random integer
in the given range. For example `"I want [7-111] apples"`.

You can use negative integers such as `[-10,10]` or even a fully negative integer
range `[-127,-34]`.

## Single value from another anonymizer

By adding `{ANONYMIZER}` anywhere in your text, it will be replaced by a random value
given by the anonymizer. For example `"My email address is {email}, what's your ?"`.

## Column value from another anonymizer

By adding `{ANONYMIZER:COLUMN}` anywhere in your text, it will be replaced by a random
value for the named column given by the anonymizer. For example, `"The country I live into is {address:country}."`.

If you use more than one column of the same anonymizer in the same text value, the same
generated row from the target anonymizer will be used. For example, this text will give
a consistent result: `"The country I live in {address:locality} in {address:country}."`.

### DIFF
--- a/docs/content/anonymization/core-anonymizers/pattern.md
+++ b/docs/content/anonymization/core-anonymizers/pattern.md
@@ -1,0 +1,81 @@
+## String pattern
+
+Anonymize by building text values using other anonymizers to fill holes in text.
+
+@@@ standalone docker
+
+```yaml [YAML]
+# db_tools.config.yaml
+anonymization:
+    default:
+        customer:
+            biography:
+                anonymizer: pattern
+                options: {value: '{firstname} {lastname} lives in {address:locality} in {address:country}, he has [2,7] cats.'}
+  #...
+```
+
+@@@
+@@@ symfony
+
+::: code-group
+```php [Attribute]
+namespace App\Entity;
+
+use Doctrine\ORM\Mapping as ORM;
+use MakinaCorpus\DbToolsBundle\Attribute\Anonymize;
+
+#[ORM\Entity()]
+#[ORM\Table(name: 'customer')]
+class Customer
+{
+    // ...
+
+    #[ORM\Column]
+    #[Anonymize(type: 'pattern', options: ['value' => '{firstname} {lastname} lives in {address:locality} in {address:country}, he has [2,7] cats.'])] // [!code ++]
+    private ?string $biography = null;
+}
+```
+
+```yml [YAML]
+# config/anonymization.yaml
+
+customer:
+    biography:
+        anonymizer: pattern
+        options: {value: '{firstname} {lastname} lives in {address:locality} in {address:country}, he has [2,7] cats.'}
+#...
+```
+:::
+
+@@@
+
+### Integer range
+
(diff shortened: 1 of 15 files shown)


docs/content/anonymization/core-anonymizers/pattern.md
```diff
@@ -0,0 +1,83 @@
+## String pattern
+
+Anonymize by building text values using other anonymizers to fill holes in text.
+
+### String range
```
SimonMellerin, line 5: maybe `Integer range`  ?

docs/content/anonymization/core-anonymizers/pattern.md
```diff
@@ -0,0 +1,83 @@
...
+You can use negative integers such as `[-10,10]` or even a fully negative integer
+range `[-127,-34]`.
+
+### Single value from another anonymizer
+
+By adding `{ANONYMIZER}` anywhere in your text, it will be replaced by a random value
+given by the anonymizer. For example `"My email address is {email}, what's your ?"`.
+
+:::warning
+The target anonymizer must not be a column anonymizer.
```
SimonMellerin, line 19:
```suggestion
The target anonymizer must not be a mutli-column anonymizer.
```

docs/content/anonymization/core-anonymizers/pattern.md
```diff
@@ -0,0 +1,83 @@
...
+:::
+
+### Column value from another anonymizer
```
SimonMellerin, line 22:
```suggestion
### Column value from a mutli-column anonymizer
```

SimonMellerin (commented):
Seems really good !

Maybe an alternative could be to accept this kind of conf

```yaml
anonymization:
    default:
        customer:
            biography:
                anonymizer: pattern
                options: 
                     value: '{1} {2} has email address {3}.'
                     1: firstname
                     2: lastname
                     3: 
                         anonymizer: email
                         options: {domain: 'custom-domain.com'}
```

This can be an addition for advanced cases. what do you think ?

But this is may be far to complex for a first version.

pounard:
> Maybe an alternative could be to accept this kind of conf

It seems vastly different from what I have done, but it's definitely an interesting idea. How about opening an issue to add this dialect variant as well ?

docs/content/anonymization/core-anonymizers/pattern.md
```diff
@@ -0,0 +1,83 @@
+## String pattern
+
+Anonymize by building text values using other anonymizers to fill holes in text.
+
```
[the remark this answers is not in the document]
pounard: OK, I'll rework the doc.

PR #240, opened 2025-11-01
feat: migrate phpstan 1 => 2

Chris53897:
#205


SimonMellerin (changes requested):

SimonMellerin:
I'm closing this issue.

Upgrading PHPStan version while ignoring all the new issues raised by this new version is not the way we want to do things.

I suggest you create an issue about this PHPStan upgrade, and then we will make some tests on our own to see which implication it may cause.

Thanks

PR #244, opened 2025-12-08
feat: update phpstan support rules with PHP 8.4 and solve errors

shakaran:
This PR solve errors with PHP 8.4 

The changes maintain backward compatibility while adhering to the new PHP 8.4 type system requirements.

it mostly update method signatures in several files to explicitly declare nullable types for parameters with null defaults


src/Bridge/Standalone/PharCompiler.php
```diff
@@ -157,9 +156,11 @@ public function compile(?string $pharFile = null): void
         unset($phar);
 
         // re-sign the phar with reproducible timestamp / signature
-        $util = new Timestamps($pharFile);
-        $util->updateTimestamps($this->versionDate);
-        $util->save($pharFile, \Phar::SHA512);
+        if (\class_exists('Seld\\PharUtils\\Timestamps')) {
```
pounard: I'm dubious about this change, the code that is now within the if statement is not supposed to be optional, it's the whole goal of this method!

I think it should be a if (not ...) throw some exception instead. It will make PHPStan happy the same way.

src/Restorer/RestorerFactory.php
```diff
@@ -47,8 +47,6 @@ public function create(?string $connectionName = null): AbstractRestorer
             $this->configRegistry->getConnectionConfig($connectionName),
         );
 
-        \assert($restorer instanceof AbstractRestorer);
```
pounard, line 50: Fact is, I often add those assert() statements due to some IDE bugs that sometime don't find the right types. Removing those will make me angry in some way. Is there a good reason removing this?

src/Bridge/Standalone/PharCompiler.php
shakaran: I add this, since seld/phar-utils was not declared implicit or transient dep in require-dev so the class was not found.

I modify now to add as require-dev dependency explicit. Fixed at e3bd809

src/Restorer/RestorerFactory.php
shakaran: Yes, since the return type is AbstractRestorer is detected already in the return, so it is infered properly by all IDEs and no needed anymore

SimonMellerin:
Thanks for this PR and your work here!

I'm sorry, my last changes for PHP-8.5 support conflict with your PR :/

Can you rebase main on your branch and solve these conflicts ?

Thanks

src/Restorer/RestorerFactory.php
pounard: ~~Mine doesn't actually...~~ forget that, it's OK indeed.

PR #245, opened 2025-12-12
Introduce a new "connection provider" for standalone usage

pounard:
In some case, static database URL is not enough, the user might want to dynamically create or fetch database URL (multi-tenant architecture, using environment variables, etc...).


pounard:
@SimonMellerin I explicitly marked and documented the feature as being experimental. On my side this is ready to be merged.

PR #251, opened 2026-01-23
no issue - simplify dev.sh and local testing docker stack, incremental github ci updates, remove mysql 5.7

pounard:
Quite a large PR, but it would be hard to split:

- simplify the `dev.sh` script by moving out all `DBAL_*` env vars,
- move out some env vars directly into a docker env ini file,
- share the same Dockerfile for all PHP versions,
- remove MySQL 5.7 automatic testing,
- add ~~PostgreSQL 18 and~~ MariaDB 12 automatic testing, EDIT: actually it does work great with PostgreSQL 18, but because of changes in `pg_dump`, tests cannot run with all versions using the same `Dockerfile`,
- incrementally fix the github CI (mostly raise some dependencies versions).


.github/workflows/continuous-integration.yml
```diff
@@ -225,12 +201,12 @@ jobs:
...
         php-version:
-          - "8.4"
+          - "8.5"
         mysql-version:
-          - "5.7"
           - "8.0"
           - "8.2"
           - "8.3"
+          - "8.4"
         extension:
```
SimonMellerin, line 210: May be this requires a documentation update

docker/php-dev/Dockerfile
```diff
@@ -0,0 +1,96 @@
+# Dockerfile for PHP 8.1 and greater.
```
SimonMellerin, line 1: I think I prefer this file to be named `docker/php-dev/Dockerfile` to keep the nomenclature clean and to explicit its purpose

SimonMellerin (changes requested):

docker/php-dev/Dockerfile
pounard: When I started I made different Dockerfile for PHP 8.1 to 8.4 and PHP 8.5 because sqlsrv driver wasn't available yet. I thought that this may happen once again, that PHP versions diverges so much that we'd need to create a new Dockerfile. But OK, until  we don't, I'll rename this file.

SimonMellerin: I would like to have this Dockerfile in its own directory like others `php-dev/Dockerfile`

pounard: Done!

SimonMellerin (approved):

